PR #9, opened 2026-03-17
feat: replayer result export, sei-config consolidation, and API cleanup

bdchatham:
## Summary
- Add `ResultExport` to `ReplayerSpec` (exclusive to replayer, not full nodes) for periodic block result export to S3
- Remove `StartHeight` from `S3SnapshotSource` — users provide an explicit S3 URI for a specific snapshot; height is auto-discovered from the archive key name
- Replace local `wellKnownSnapshotChains` map with `seiconfig.KnownChain()` and local port constants with `seiconfig.PortRPC`, `seiconfig.PortSidecar`, `seiconfig.NodePorts()`
- Use explicit task status from sidecar (`running`/`completed`/`failed`) instead of inferring from `CompletedAt`
- Update sample manifest to `pacific-1-shadow-replayer.yaml` with inferred S3 bucket
- Bump `sei-config` to v0.0.7, `seictl` to v0.0.15
- Remove stale `conditions.go` and old `pacific-1-replay.yaml` sample

## Test plan
- [x] Replayer validation tests (well-known chain, custom chain requires URI, peer requirements)
- [x] BuildTask tests (inferred bucket, explicit URI, discover peers, result export)
- [x] Result export task builder tests (replayer-only, nil when absent)
- [x] Resource generation tests updated for sei-config port constants
- [x] Plan execution tests updated for explicit task status model
- [x] CI lint and test pass


internal/controller/node/task_builders.go
```diff
@@ -7,8 +7,12 @@ import (
 )
 
 const (
-	defaultSnapshotUploadCron = "0 0 * * *"
-	defaultSnapshotInterval   = int64(2000)
+	defaultSnapshotUploadCron  = "0 0 * * *"
+	defaultSnapshotInterval    = int64(2000)
+
+	resultExportBucket = "sei-node-mvp"
```
bdchatham, line 13: These will be updated to a real bucket once MVP is complete.

PR #34, opened 2026-03-29
feat: blue-green deployment orchestration for SeiNodeGroup

bdchatham:
## Summary

- Adds `UpdateStrategy` to `SeiNodeGroupSpec` with two strategies: `BlueGreen` (deploy when green nodes catch up to tip) and `HardFork` (deploy at a specific block height via sidecar SIGTERM)
- Deployment orchestration lives inside SeiNodeGroup — single owner of all SeiNodes, no separate CRD
- Reuses existing `TaskPlan` / `PlanExecutor` machinery with 7 new controller-side task types
- Gates `ensureSeiNode` during active deployments to prevent blue node mutation
- Adds revision labels (`sei.io/revision`) for Service selector targeting during traffic switch
- Removes `SeiNodeStorageConfig` / `RetainOnDelete` (simplifies API)

### HardFork task sequence
`create-green-nodes` → `await-green-running` → `submit-halt-signal-blue` → `await-green-at-height` → `switch-traffic` → `teardown-blue`

### BlueGreen task sequence
`create-green-nodes` → `await-green-running` → `await-green-caught-up` → `switch-traffic` → `teardown-blue`

### Key design decisions
- No self-mutation: user's spec change is the trigger, controller never writes to its own spec
- Generation-based revision labels (not "blue"/"green" strings)
- On failure: mark Degraded, keep blue running, no auto-rollback
- Concurrent deployments rejected: active plan must complete first
- halt-height uses sidecar `await-condition(height=H, action=SIGTERM)` — no seid config patching needed

## Test plan

- [x] All existing tests pass
- [x] Zero lint issues
- [ ] Integration test: HardFork deployment with halt-height
- [ ] Integration test: BlueGreen deployment with catching_up signal
- [ ] Test concurrent deployment rejection
- [ ] Test failure handling (green node fails → Degraded, blue stays)


api/v1alpha1/seinodegroup_types.go
```diff
@@ -41,6 +41,49 @@ type SeiNodeGroupSpec struct {
...
+	// block height. The old binary halts via sidecar SIGTERM at the
+	// configured halt-height, and the new binary continues from that height.
+	UpdateStrategyHardFork UpdateStrategyType = "HardFork"
+)
+
+// UpdateStrategy controls how spec changes propagate to child SeiNodes.
+// +kubebuilder:validation:XValidation:rule="self.type != 'HardFork' || (has(self.hardFork) && self.hardFork.haltHeight > 0)",message="hardFork strategy requires haltHeight > 0"
+type UpdateStrategy struct {
+	// Type selects the deployment strategy.
+	Type UpdateStrategyType `json:"type"`
```
bdchatham, line 72: Are we missing BlueGreen type here?

internal/controller/nodegroup/controller.go
```diff
@@ -80,6 +80,30 @@ func (r *SeiNodeGroupReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+	// Detect if a new deployment is needed: generation changed + strategy set.
+	if planner.NeedsDeployment(group) {
+		result, err := r.startDeployment(ctx, group, statusBase)
+		if err != nil {
+			logger.Error(err, "starting deployment")
+			observability.ReconcileErrorsTotal.WithLabelValues(controllerName, ns, name).Inc()
+			return ctrl.Result{}, fmt.Errorf("starting deployment: %w", err)
+		}
+		return result, nil
+	}
```
bdchatham: I'm thinking this could be better abstracted or generalized. Seems like this could all be handled behind a single r.reconcileDeployment() method. Does there need to be an explicit difference between reconcile and start?

internal/controller/nodegroup/deployment.go
```diff
@@ -0,0 +1,146 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	seiv1alpha1 "github.com/sei-protocol/sei-k8s-controller/api/v1alpha1"
+	"github.com/sei-protocol/sei-k8s-controller/internal/planner"
+)
+
+// startDeployment detects a pending deployment, builds the plan, and
+// writes the initial DeploymentStatus. The next reconcile will pick
+// up the active plan via reconcileDeployment.
```
bdchatham: I think this sort-of points it out well. If it's identified that a new deployment is observed, then simply do this work as part of reconciliation and put the deployment in active state, then continue to reconciliation. Putting this behind a condition makes sense, but it is conceptually part of deployment reonciliation.

internal/planner/deployment.go
```diff
@@ -0,0 +1,202 @@
...
+			return nil, err
+		}
+		tasks[i] = t
+	}
+	return tasks, nil
+}
+
+// buildPlannedTask2 is like buildPlannedTask but uses the group name
+// instead of node name for deterministic IDs.
+func buildPlannedTask2(resourceName, taskType string, attempt int, params any) (seiv1alpha1.PlannedTask, error) {
```
bdchatham: This should have a better name that suggests it uses group name.

internal/task/deployment_await.go
```diff
@@ -0,0 +1,258 @@
...
+	sidecar "github.com/sei-protocol/seictl/sidecar/client"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	seiv1alpha1 "github.com/sei-protocol/sei-k8s-controller/api/v1alpha1"
+)
+
+// ---------------------------------------------------------------------------
+// await-green-running: polls until all green SeiNodes reach Running phase.
+// ---------------------------------------------------------------------------
```
bdchatham: I'm okay with descriptive comments that describe complexity not intuitive to the local code, but I'd suggests not using this header/page break style

internal/planner/deployment.go
```diff
@@ -0,0 +1,202 @@
...
+
+// IsDeploymentActive returns true when a deployment plan is in progress.
+func IsDeploymentActive(group *seiv1alpha1.SeiNodeGroup) bool {
+	return group.Status.Deployment != nil &&
+		group.Status.Deployment.Plan.Phase == seiv1alpha1.TaskPlanActive
+}
+
+// BuildDeploymentPlan builds the task plan for a blue-green deployment
+// based on the configured update strategy.
+func BuildDeploymentPlan(
```
bdchatham: This is a problem that it doesn't meet the generic planner interface. We need to store the blue node (let's use something less specific to deployment naming, something like incumbentNodes and entrantNodes). If we can store the incumbentNodes on the SeiNodeGroup, you can keep alignment with the interface which will be very powerful in the reconciliation complexity.

internal/task/deployment_await.go
```diff
@@ -0,0 +1,258 @@
...
+	for _, name := range e.params.NodeNames {
+		node := &seiv1alpha1.SeiNode{}
+		if err := e.cfg.KubeClient.Get(ctx, types.NamespacedName{Name: name, Namespace: e.params.Namespace}, node); err != nil {
+			return fmt.Errorf("getting green node %s: %w", name, err)
+		}
+		sc, err := buildSidecarClientForNode(e.cfg, node)
+		if err != nil {
+			return fmt.Errorf("building sidecar client for %s: %w", name, err)
+		}
+		taskID := uuid.NewSHA1(uuid.MustParse("b7e89c3a-4f12-4d8b-9a6e-1c2d3e4f5a6b"),
```
bdchatham: nit: make a var for this

internal/task/deployment_await.go
```diff
@@ -0,0 +1,258 @@
...
+	return &awaitGreenAtHeightExecution{
+		id:        id,
+		params:    p,
+		cfg:       cfg,
+		status:    ExecutionRunning,
+		submitted: make(map[string]uuid.UUID),
+	}, nil
+}
+
+func (e *awaitGreenAtHeightExecution) Execute(ctx context.Context) error {
```
bdchatham: nit: I'd suggest node using Green or Blue in names unless something is explicitly scoped to BlueGreen like a deployment strategy name. I'd argue this is just waiting for the node to reach a specified height

internal/task/deployment_await.go
```diff
@@ -0,0 +1,258 @@
...
+		if _, err := sc.SubmitTask(ctx, req); err != nil {
+			logger.Info("failed to submit await-height to green node, will retry", "node", name, "error", err)
+			return err
+		}
+		e.submitted[name] = taskID
+	}
+	return nil
+}
+
+func (e *awaitGreenAtHeightExecution) Status(ctx context.Context) ExecutionStatus {
```
bdchatham: This method (and some others) are quite long. I'd suggest adding helper methods to make the code more self-documenting as to what is actually happening and making the flow intuitive.

internal/task/deployment_create.go
```diff
@@ -0,0 +1,98 @@
...
+		e.err = err
+		return err
+	}
+
+	for i, name := range e.params.NodeNames {
+		// Check if already exists (idempotent).
+		existing := &seiv1alpha1.SeiNode{}
+		if err := e.cfg.KubeClient.Get(ctx, types.NamespacedName{Name: name, Namespace: e.params.Namespace}, existing); err == nil {
+			continue
+		}
```
bdchatham: It makes sense that we're creating nodes like this, but say we do this, does the StatefulSet adopt these instances and would it be simpler to create an entrant statefulset that is switched to once it's ready?

internal/task/deployment_switch.go
```diff
@@ -0,0 +1,136 @@
...
+		e.status = ExecutionFailed
+		e.err = fmt.Errorf("no deployment status on group %s", e.params.GroupName)
+		return e.err
+	}
+
+	// The plan executor patches status after each task, and the group
+	// controller uses activeRevision in groupSelector(). Updating
+	// activeRevision here causes the next Service reconciliation to
+	// point traffic at the green nodes.
+	group.Status.Deployment.ActiveRevision = e.params.IncomingRevision
```
bdchatham: I like this method but would love to understand the Deployment revision mechanism more deeply here. Are we tracking deployment revisions/history?

internal/task/deployment_switch.go
```diff
@@ -0,0 +1,136 @@
...
+		params: p,
+		cfg:    cfg,
+		status: ExecutionRunning,
+	}, nil
+}
+
+func (e *teardownBlueExecution) Execute(ctx context.Context) error {
+	logger := log.FromContext(ctx)
+
+	for _, name := range e.params.NodeNames {
```
bdchatham, line 88: Same point as above - operating on the statefulset seems like it would be more idiomatic/ergonomic

internal/planner/deployment.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	return strconv.FormatInt(group.Generation-1, 10)
+}
+
+// hardForkDeploymentPlanner builds a deployment plan for the HardFork strategy.
+type hardForkDeploymentPlanner struct{}
+
+func (p *hardForkDeploymentPlanner) BuildPlan(
+	group *seiv1alpha1.SeiNodeGroup,
+	_ []seiv1alpha1.SeiNode,
```
bdchatham, line 70: Need to delete this so it matches the interface.

PR #35, opened 2026-03-30
refactor: generalize SeiNodeGroup reconcile loop with unified plan orchestration

bdchatham:
## Summary

- Restructures the reconcile loop into a clean sequence: `reconcileSeiNodes` → `reconcilePlan` → `reconcileNetworking` → `reconcileMonitoring` → `updateStatus`
- `reconcileSeiNodes` owns its own gating via `PlanInProgress` condition — any active plan (genesis, deployment) pauses node mutations
- `reconcilePlan` is the single generic entry point for all plan-driven orchestration, replacing both `reconcileGenesisAssembly` and `reconcileDeployment`
- Genesis finalization (peer collection + distribution) moves into a `collect-and-set-peers` task in the genesis plan
- Genesis planner reads `IncumbentNodes` from group status instead of requiring a `nodes` param
- `ConditionDeploymentInProgress` → `ConditionPlanInProgress` (generic across all plan types)
- Removes `deployment.go` (superseded by `plan.go`), strips `genesis.go` to utility functions

## Test plan

- [x] All existing tests pass
- [x] Zero lint issues
- [ ] Integration test: genesis ceremony with new collect-and-set-peers task
- [ ] Verify PlanInProgress condition gates node mutations during active plans


internal/controller/nodegroup/plan.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	// No active plan — check if one needs to be created.
+	if needsGenesisPlan(group) {
+		return r.startGenesisPlan(ctx, group, statusBase)
+	}
+
+	if planner.NeedsDeployment(group) {
+		return r.startDeploymentPlan(ctx, group, statusBase)
+	}
```
bdchatham: I would suggest consolidating this all into a single Plan. We can rename InitPlan to Plan. When the reconciliation identifies a condition, that condition should be checked against our planner to generate a plan. That plan should go into the Plan variable. The Plan should iteratively be making progress through this reconciliation. We don't need distinction between deployment vs genesis plan reconciliation. Just generate the plan struct and and call status/execute on the current task.

internal/controller/nodegroup/plan.go
```diff
@@ -0,0 +1,191 @@
...
+	if err := r.Status().Patch(ctx, group, statusBase); err != nil {
+		return ctrl.Result{}, fmt.Errorf("writing genesis plan: %w", err)
+	}
+
+	r.Recorder.Event(group, corev1.EventTypeNormal, "GenesisAssemblyStarted", "Starting genesis assembly plan")
+	return planner.ResultRequeueImmediate, nil
+}
+
+func (r *SeiNodeGroupReconciler) startDeploymentPlan(ctx context.Context, group *seiv1alpha1.SeiNodeGroup, statusBase client.Patch) (ctrl.Result, error) {
+	dp, err := planner.ForDeployment(group)
```
bdchatham: I would suggest rolling this logic into the planner.ForGroup

internal/controller/nodegroup/plan.go
```diff
@@ -0,0 +1,191 @@
...
+	entrantRevision := planner.EntrantRevision(group)
+	incumbentRevision := planner.IncumbentRevision(group)
+
+	group.Status.Deployment = &seiv1alpha1.DeploymentStatus{
+		Plan:              *plan,
+		IncumbentRevision: incumbentRevision,
+		EntrantRevision:   entrantRevision,
+		EntrantNodes:      planner.EntrantNodeNames(group),
+	}
+	group.Status.Phase = seiv1alpha1.GroupPhaseUpgrading
```
bdchatham: I would suggest making this part of the reconcileGroup step. In plain english it says "It seems a new deployment is being triggered based on the spec diff. We should update the status based on this, informing all following reconciliation steps that there is a deployment which is derived by the condition and state I am setting here." If you do this, all of the scenario specific logic goes into the spec reconciliation early in the code to derive a condition. That condition drives downstream outcomes like planner selection and plan creation.

internal/planner/group.go
```diff
@@ -18,13 +18,14 @@ type genesisGroupPlanner struct{}
 //     they picked up the genesis.
 func (p *genesisGroupPlanner) BuildPlan(
 	group *seiv1alpha1.SeiNodeGroup,
-	nodes []seiv1alpha1.SeiNode,
+	_ []seiv1alpha1.SeiNode,
```
bdchatham: We should just fix the interface and get rid of this parameter. It's no longer being used.

internal/task/genesis_peers.go
```diff
@@ -0,0 +1,125 @@
...
+	return nil
+}
+
+func (e *collectAndSetPeersExecution) fail(err error) error {
+	e.status = ExecutionFailed
+	e.err = err
+	return err
+}
+
+func (e *collectAndSetPeersExecution) Status(_ context.Context) ExecutionStatus { return e.status }
```
bdchatham: This should check if all of the tasks completed on the sidecar endpoints they were submitted to.

internal/controller/nodegroup/controller.go
```diff
@@ -100,19 +85,17 @@ func (r *SeiNodeGroupReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
-				return ctrl.Result{}, fmt.Errorf("updating status: %w", err)
-			}
-			return result, nil
+	planResult, planErr := r.reconcilePlan(ctx, group, statusBase)
+	if planErr != nil {
+		logger.Error(planErr, "reconciling plan")
+		observability.ReconcileErrorsTotal.WithLabelValues(controllerName, ns, name).Inc()
+		return ctrl.Result{}, fmt.Errorf("reconciling plan: %w", planErr)
+	}
+	if planResult.RequeueAfter > 0 || planResult.Requeue {
```
bdchatham: I'd add a shouldRequeue() method that we call here instead of checking those directly.

PR #36, opened 2026-03-30
fix: address expert review findings

bdchatham:
## Summary

Fixes all critical and medium-priority issues identified by the kubernetes-specialist and platform-engineer expert review.

### Critical fixes (would cause incorrect behavior)
- **Plan/Deployment status never cleared** — `completePlan` and `failPlan` now nil out `Plan` and `Deployment`, unblocking subsequent deployments
- **Phase overwrite** — `updateStatus` preserves `Upgrading` phase when `Deployment` is non-nil
- **Revision label not on pods** — `sei.io/revision` added to `PodLabels` in both `generateSeiNode` and `ensureEntrantNode`, making Service selector work
- **Unreliable incumbent revision** — uses `ObservedGeneration` instead of `Generation-1`
- **Post-deployment naming divergence** — `ensureSeiNode` now uses `findNodeByOrdinal` (label-based lookup) instead of name-based, so it finds entrant-named nodes

### High priority fixes
- `detectDeploymentNeeded` gated behind `PlanInProgress` and `ObservedGeneration > 0`
- `Upgrading` added to `allGroupPhases` in metrics
- `ObservedGeneration` updates whenever no plan is active (not gated on `updateStrategy`)
- Deduplicated name/revision computation via planner exports

### Medium priority fixes
- `switchTrafficExecution` patches via KubeClient (not in-memory mutation)
- `createEntrantNodes` distinguishes transient Get errors from NotFound
- `needsGenesisPlan` uses constants not string literals
- `completePlan` returns networking errors instead of swallowing
- `collectAndSetPeers` Status() trusts Execute completion

## Test plan
- [x] All existing tests pass
- [x] Zero lint issues

### DIFF
--- a/internal/controller/nodegroup/nodes.go
+++ b/internal/controller/nodegroup/nodes.go
@@ -69,7 +69,6 @@ func (r *SeiNodeGroupReconciler) detectDeploymentNeeded(group *seiv1alpha1.SeiNo
 		EntrantRevision:   planner.EntrantRevision(group),
 		EntrantNodes:      planner.EntrantNodeNames(group),
 	}
-	group.Status.Phase = seiv1alpha1.GroupPhaseUpgrading
 }
 
 // populateIncumbentNodes lists child SeiNodes and records their names
--- a/internal/controller/nodegroup/nodes.go
+++ b/internal/controller/nodegroup/nodes.go
@@ -5,17 +5,16 @@ import (
 	"fmt"
 	"maps"
 	"slices"
-	"strconv"
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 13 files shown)


internal/controller/nodegroup/nodes.go
```diff
@@ -54,22 +53,20 @@ func (r *SeiNodeGroupReconciler) detectDeploymentNeeded(group *seiv1alpha1.SeiNo
 	if group.Generation == group.Status.ObservedGeneration {
 		return
 	}
+	if group.Status.ObservedGeneration == 0 {
+		return // new group, never reconciled yet
+	}
```
bdchatham, line 59: Can we add a similar, but opposite, check for this to gate genesis? Genesis should only ever happen on the first generation

PR #37, opened 2026-03-30
fix: decouple Service selector from revision, add template hash for change classification

bdchatham:
## Summary

Fixes two gaps in the deployment model identified by expert review:

### 1. Scaling no longer breaks traffic
- `groupSelector` excludes `sei.io/revision` during steady state — only includes it during active deployments for traffic pinning between incumbent/entrant sets
- `groupOnlySelector` used for `listChildSeiNodes`/`scaleDown` where all children must be visible
- A replicas change bumps generation but does NOT change the template hash, so no deployment triggers

### 2. Sidecar upgrades are in-place, not blue-green
- `TemplateHash` on status hashes only deployment-worthy fields: `image`, `entrypoint`, `chainId`
- `detectDeploymentNeeded` compares hash instead of generation
- Sidecar image, overrides, and podLabels changes propagate in-place via `ensureSeiNode` → SeiNode controller → StatefulSet SSA

### Change classification matrix

| Change | Hash changes? | Result |
|--------|:-:|--------|
| `spec.template.spec.image` | Yes | Blue-green deployment |
| `spec.template.spec.entrypoint` | Yes | Blue-green deployment |
| `spec.template.spec.chainId` | Yes | Blue-green deployment |
| `spec.template.spec.sidecar.image` | No | In-place rolling restart |
| `spec.template.spec.overrides` | No | In-place config update |
| `spec.replicas` | No | Scale operation |

### 3. Node naming reverted to `{group}-{ordinal}`
- Generation-qualified names only for deployment entrant nodes
- Matches Kubernetes StatefulSet naming convention

## Test plan
- [x] All existing tests pass
- [x] Zero lint issues

### DIFF
--- a/internal/controller/nodegroup/labels.go
+++ b/internal/controller/nodegroup/labels.go
@@ -1,9 +1,12 @@
 package nodegroup
 
 import (
+	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
 	"maps"
 	"strconv"
+	"strings"
 
 	seiv1alpha1 "github.com/sei-protocol/sei-k8s-controller/api/v1alpha1"
 )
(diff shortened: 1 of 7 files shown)


internal/controller/nodegroup/labels.go
```diff
@@ -83,3 +96,17 @@ func resourceLabels(group *seiv1alpha1.SeiNodeGroup) map[string]string {
 func managedByAnnotations() map[string]string {
 	return map[string]string{managedByAnnotation: controllerName}
 }
+
+// templateHash computes a hash over spec fields that require new nodes
+// when changed. Fields that can be updated in-place on existing nodes
+// (sidecar, podLabels, overrides) are excluded.
+func templateHash(spec *seiv1alpha1.SeiNodeSpec) string {
+	h := sha256.New()
+	h.Write([]byte(spec.Image))
```
bdchatham, line 104: I would only hash this value. ChainID should never change and not be respected if it does. Entry point and args are important but I'd leave them out for now

PR #46, opened 2026-04-01
feat: chain fork support as genesis ceremony extension

bdchatham:
## Summary

Minimal fork support for SeiNodeGroup — fork an existing chain's state into a private network.

### Design
Fork is an extension of the existing genesis ceremony, not a separate flow. The genesis planner selects the assembler task based on which condition is set:
- `GenesisCeremonyNeeded` → `assemble-and-upload-genesis` (existing)
- `ForkGenesisCeremonyNeeded` → `assemble-fork-genesis` (new sidecar task)

### CRD
```go
type ForkConfig struct {
    SourceChainID string `json:"sourceChainId"`
}
```

Added as `Fork *ForkConfig` on `GenesisCeremonyConfig`. That's it — the exported state location and assembled genesis output use the existing platform S3 conventions.

### Example
```yaml
genesis:
  chainId: private-fork-1
  fork:
    sourceChainId: pacific-1
```

### Changes
- `ForkConfig` type + `Fork` field on `GenesisCeremonyConfig`
- `GenesisCeremonyNeeded` and `ForkGenesisCeremonyNeeded` conditions
- `detectGenesisCeremonyNeeded` replaces separate detection methods
- `genesisGroupPlanner` branches on condition to select assembler task
- `AssembleForkGenesisParams` + Deserialize registration

### Companion
Seictl `assemble-fork-genesis` handler (PR coming)


internal/task/task.go
```diff
@@ -220,6 +220,10 @@ func Deserialize(taskType, id string, params json.RawMessage, cfg ExecutionConfi
 	case TaskTypeTeardownNodes:
 		return deserializeTeardownNodes(id, params, cfg)
 
+	// Fork genesis assembly (sidecar task)
```
bdchatham: nit: not necessary to specify it being a sidecar task.

internal/controller/nodegroup/nodes.go
```diff
@@ -41,9 +41,42 @@ func (r *SeiNodeGroupReconciler) reconcileSeiNodes(ctx context.Context, group *s
...
 	r.detectDeploymentNeeded(group)
+	r.detectGenesisCeremonyNeeded(group)
 	return nil
 }
 
+// detectGenesisCeremonyNeeded sets either GenesisCeremonyNeeded or
+// ForkGenesisCeremonyNeeded depending on whether the group has a fork
+// config. The genesis planner reads this condition to select the
+// appropriate assembler task.
+func (r *SeiNodeGroupReconciler) detectGenesisCeremonyNeeded(group *seiv1alpha1.SeiNodeGroup) {
```
bdchatham, line 52: Are we still enforcing genesis can only be set when generation is 0? If yes, then we may not ever enter this code a second time and in that case, we need to unset the conditions elsewhere. 

Let's discuss this with the Kubernetes expert.

internal/planner/group.go
```diff
@@ -7,7 +7,10 @@ import (
 	"github.com/sei-protocol/sei-k8s-controller/internal/task"
 )
 
-const groupAssemblyMaxRetries = 60
+const (
+	groupAssemblyMaxRetries = 60
+	TaskAssembleForkGenesis = "assemble-fork-genesis"
```
bdchatham: Let's source this from seictl. The name is now "assemble-genesis-fork"

internal/controller/nodegroup/plan.go
```diff
@@ -82,6 +82,12 @@ func (r *SeiNodeGroupReconciler) completePlan(ctx context.Context, group *seiv1a
 		group.Status.Deployment = nil
 	}
 
+	// Genesis/fork ceremony finalization.
+	if group.Spec.Genesis != nil && group.Status.Deployment == nil {
+		setCondition(group, seiv1alpha1.ConditionGenesisCeremonyComplete, metav1.ConditionTrue,
+			"CeremonyComplete", "Genesis ceremony completed")
+	}
+
```
bdchatham: Is this saying "when genesis is provided and a deployment is identified, mark genesis complete"?

internal/planner/group.go
```diff
@@ -7,7 +7,10 @@ import (
 	"github.com/sei-protocol/sei-k8s-controller/internal/task"
 )
 
-const groupAssemblyMaxRetries = 60
+const (
+	groupAssemblyMaxRetries = 60
+	TaskAssembleGenesisFork = "assemble-genesis-fork"
```
bdchatham: Let's take this from the seictl version

internal/task/fork_export.go
```diff
@@ -0,0 +1,211 @@
...
+func (e *submitExportStateExecution) Status(ctx context.Context) ExecutionStatus {
+	if s, done := e.isTerminal(); done {
+		return s
+	}
+	if !e.submitted {
+		return ExecutionRunning
+	}
+	// TODO: Poll sidecar for export task completion.
+	// For now, return Running (the task will be re-submitted on restart).
+	return ExecutionRunning
```
bdchatham: We need to fix these now that seictl changes have landed, right?

internal/planner/planner.go
```diff
@@ -72,36 +71,47 @@ func ForGroup(group *seiv1alpha1.SeiNodeGroup) (GroupPlanner, error) {
...
-	}
 	if group.Status.Plan != nil {
 		return false
 	}
+	genesisNeeded := hasCondition(group, seiv1alpha1.ConditionGenesisCeremonyNeeded)
+	forkNeeded := hasCondition(group, seiv1alpha1.ConditionForkGenesisCeremonyNeeded)
+	if !genesisNeeded && !forkNeeded {
+		return false
+	}
+	return int32(len(group.Status.IncumbentNodes)) >= group.Spec.Replicas
```
bdchatham, line 92: Can we make this into a method that returns bool so that we can describe the check with the method name rather than leaving it up to the reader to know?

PR #51, opened 2026-04-03
feat: sidecar probes, SubmittedAt diagnostics, submission logging

bdchatham:
## Summary

Prod hardening: wire the sidecar's health endpoints to Kubernetes probes, add operator diagnostics for stuck nodes, and improve log correlation.

### Changes

- **Liveness probe** on sidecar container → `GET /v0/livez` (SQLite health). Kubelet restarts the pod if the sidecar's backing store is wedged.
- **Readiness probe** on sidecar container → `GET /v0/healthz` (mark-ready gate). Pod is removed from Service endpoints until bootstrap completes. FailureThreshold=6 to accommodate slow bootstrap.
- **`SubmittedAt` on PlannedTask** — set on first successful sidecar submission, nil before. Lets operators distinguish "never submitted" from "submitted 30 minutes ago" when diagnosing stuck nodes. Cleared on retry so retried tasks get a fresh timestamp.
- **Info-level log on task submission** — `"task submitted to sidecar"` with task type and ID for controller↔sidecar log correlation.

## Test plan

- [x] `make test` — all existing tests pass
- [x] `make lint` — 0 issues
- [x] `make manifests generate` — CRD YAML and DeepCopy regenerated


internal/controller/node/resources.go
```diff
@@ -124,6 +124,28 @@ func buildSidecarContainer(node *seiv1alpha1.SeiNode, platform PlatformConfig) c
...
+			ProbeHandler: corev1.ProbeHandler{
+				HTTPGet: &corev1.HTTPGetAction{
+					Path: "/v0/healthz",
+					Port: intstr.FromInt32(port),
+				},
+			},
+			InitialDelaySeconds: 2,
+			PeriodSeconds:       5,
+			FailureThreshold:    6,
+		},
```
bdchatham, line 148: Will these be okay for our workflows that take a while to setup the node, like snapshot restore or will the controller kill nodes that are still healthy just taking a while to setup?

bdchatham: The liveness probe (`/v0/livez`) only checks SQLite health — it does not care about bootstrap progress. A node doing a 2-hour snapshot restore will pass livez the entire time because the sidecar process is alive and the store is responsive.

The readiness probe (`/v0/healthz`) will return 503 during bootstrap, which is correct — it means "don't route traffic to this pod yet." But readiness failures never cause restarts, they only remove the pod from Service endpoints.

So: long bootstraps will NOT trigger restarts. The only thing that restarts the pod is a failed livez (SQLite wedged/corrupted), which is genuinely irrecoverable.

PR #63, opened 2026-04-06
feat: config drift detection and runtime plan execution for SeiNode

bdchatham:
## Summary

- **Replace field-by-field spec comparison** in `ensureSeiNode` with `equality.Semantic.DeepEqual` — all SeiNodeGroup template spec changes now propagate to child SeiNode CRs automatically, eliminating the "forgotten field" class of bugs
- **Add `ConfigApplyParams` to `NodePlanner` interface** — extracts config-apply param building from each mode planner (`full`, `archive`, `validator`, `replayer`) for reuse by runtime reconfiguration plans
- **Add `BuildConfigUpdatePlan`** — produces a lightweight `discover-peers → config-apply (incremental) → config-validate` task plan for running nodes
- **Add config drift detection** — `detectConfigDrift` compares current peer discovery params against a `lastAppliedPeerParams` snapshot stored in SeiNode status; sets `ConfigUpdateNeeded` condition when drift is detected on a Running node with no active plan
- **Add plan awareness to `reconcileRunning`** — drives active plans, builds plans from conditions, then falls through to existing monitor task path; plans use the same `PlanExecutor` as init plans with no "runtime" vs "init" distinction

## Design

The SeiNodeGroup controller propagates template spec changes to child SeiNode CRs. The SeiNode controller detects changes to its own spec, sets conditions, and builds task plans to resolve them. Plans serialize — only one occupies `status.plan` at a time. Once complete and cleared, the controller checks for remaining conditions and builds the next plan.

```
spec change → condition → plan → tasks → condition cleared
```

## Test plan

- [x] `TestDetectConfigDrift_NoDrift` — no condition set when params match
- [x] `TestDetectConfigDrift_PeersAdded` — condition set when peers added to spec
- [x] `TestDetectConfigDrift_SkippedWhenPlanActive` — detection skipped during active plan
- [x] `TestDetectConfigDrift_SkippedWhenNotRunning` — detection skipped for non-Running nodes
- [x] `TestDetectConfigDrift_ClearsConditionOnRevert` — condition cleared when spec reverted
- [x] `TestPeerParamsEqual` — nil/non-nil/equal comparison
- [x] `TestBuildConfigUpdatePlan_WithPeers` — 3-task plan with discover-peers
- [x] `TestBuildConfigUpdatePlan_NoPeers` — 2-task plan without discover-peers
- [x] `TestBuildConfigUpdatePlan_ConfigApplyIsIncremental` — incremental flag set, original params not mutated
- [x] All existing tests pass (0 regressions)

### DIFF
--- a/api/v1alpha1/seinode_types.go
+++ b/api/v1alpha1/seinode_types.go
@@ -248,12 +248,6 @@ type SeiNodeStatus struct {
 	// +optional
 	ResolvedPeers []string `json:"resolvedPeers,omitempty"`
 
-	// LastAppliedPeerParams is the serialized DiscoverPeersParams from the
-	// last successful peer configuration. Compared against the current spec
-	// to detect peer config drift on Running nodes.
-	// +optional
-	LastAppliedPeerParams *apiextensionsv1.JSON `json:"lastAppliedPeerParams,omitempty"`
-
 	// ConfigStatus reports the observed configuration state from the sidecar.
 	// +optional
 	ConfigStatus *ConfigStatus `json:"configStatus,omitempty"`
--- a/api/v1alpha1/seinode_types.go
+++ b/api/v1alpha1/seinode_types.go
@@ -221,6 +221,12 @@ type SeiNodeStatus struct {
 	// Phase is the high-level lifecycle state.
 	Phase SeiNodePhase `json:"phase,omitempty"`
 
+	// ObservedGeneration is the most recent spec generation the controller
+	// has fully reconciled. Compared against metadata.generation to detect
+	// spec drift on Running nodes that may require a reconfiguration plan.
+	// +optional
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+
 	// +listType=map
 	// +listMapKey=type
 	// +optional
@@ -241,6 +247,12 @@ type SeiNodeStatus struct {
 	// future peer-update plans can detect drift.
 	// +optional
 	ResolvedPeers []string `json:"resolvedPeers,omitempty"`
+
+	// LastAppliedPeerParams is the serialized DiscoverPeersParams from the
+	// last successful peer configuration. Compared against the current spec
+	// to detect peer config drift on Running nodes.
+	// +optional
+	LastAppliedPeerParams *apiextensionsv1.JSON `json:"lastAppliedPeerParams,omitempty"`
 
 	// ConfigStatus reports the observed configuration state from the sidecar.
 	// +optional
--- a/internal/controller/node/controller.go
+++ b/internal/controller/node/controller.go
@@ -89,8 +89,8 @@ func (r *SeiNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, fmt.Errorf("reconciling peers: %w", err)
 	}
 
-	if err := r.detectConfigDrift(ctx, node); err != nil {
-		return ctrl.Result{}, fmt.Errorf("detecting config drift: %w", err)
+	if err := r.detectPeerDrift(ctx, node); err != nil {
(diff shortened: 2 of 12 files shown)


api/v1alpha1/seinode_types.go
```diff
@@ -242,6 +248,12 @@ type SeiNodeStatus struct {
 	// +optional
 	ResolvedPeers []string `json:"resolvedPeers,omitempty"`
 
+	// LastAppliedPeerParams is the serialized DiscoverPeersParams from the
+	// last successful peer configuration. Compared against the current spec
+	// to detect peer config drift on Running nodes.
+	// +optional
+	LastAppliedPeerParams *apiextensionsv1.JSON `json:"lastAppliedPeerParams,omitempty"`
```
bdchatham: This seems like an anti-pattern compared to the existing flow. Right now we consider peers as stateless. We may need to track peers at the controller level so we can patch differences automatically as they are observed, but for now, this seems to conflate the two methods.

It is possible that we should just bite the bullet and move peer reconciliation into the controller and let it push config changes as needed.

internal/controller/node/controller.go
```diff
@@ -177,6 +203,89 @@ func (r *SeiNodeReconciler) reconcileRunning(ctx context.Context, node *seiv1alp
...
+// reconfiguration plan from the current spec.
+func (r *SeiNodeReconciler) buildConfigUpdatePlan(ctx context.Context, node *seiv1alpha1.SeiNode) (ctrl.Result, error) {
+	p, err := planner.ForNode(node)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("resolving planner for config update: %w", err)
+	}
+
+	plan, err := planner.BuildConfigUpdatePlan(node, p.ConfigApplyParams(node))
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("building config update plan: %w", err)
```
bdchatham: planner.ForNode should return the config update plan. It shouldn't be a new code path. I'd suggest not calling it config update plan but something more scoped to the condition/domain it's addressing.

internal/controller/node/controller.go
```diff
@@ -177,6 +203,89 @@ func (r *SeiNodeReconciler) reconcileRunning(ctx context.Context, node *seiv1alp
...
+	})
+
+	if err := r.Status().Patch(ctx, node, patch); err != nil {
+		return ctrl.Result{}, fmt.Errorf("updating status after plan failure: %w", err)
+	}
+
+	r.Recorder.Eventf(node, corev1.EventTypeWarning, "ConfigUpdateFailed", failMsg)
+	logger.Info("config update plan failed", "detail", failMsg)
+	return ctrl.Result{}, nil
+}
```
bdchatham: Feels like this shouldn't be net-new but maybe I'm missing something.

internal/controller/node/reconfigure.go
```diff
@@ -0,0 +1,97 @@
...
+	ReasonConfigDriftDetected   = "ConfigDriftDetected"
+	ReasonConfigUpdateComplete  = "ConfigUpdateComplete"
+	ReasonConfigUpdateFailed    = "ConfigUpdateFailed"
+)
+
+// detectConfigDrift compares the current peer discovery params against the
+// last-applied snapshot stored in status. When drift is detected on a
+// Running node with no active plan, the ConfigUpdateNeeded condition is set,
+// signaling reconcileRunning to build a reconfiguration plan.
+func (r *SeiNodeReconciler) detectConfigDrift(ctx context.Context, node *seiv1alpha1.SeiNode) error {
```
bdchatham: I think config drift is too opaque. We don't want to operate at the config level of the node at this point. We are orchestrating known reconciliation for a known set of possible spec diffs

PR #69, opened 2026-04-09
refactor: derive Gateway routes from node mode with listener-based hostnames

bdchatham:
## Summary

Replace `baseDomain`/`hostnames` CRD fields with mode-derived route generation. The Gateway owns the protocol topology — deployments just attach to it.

## What Changed

**CRD simplification** — `GatewayRouteConfig` reduced to just `annotations`. No more `baseDomain` or `hostnames` fields. Gateway config just needs to exist (non-nil) to enable routing.

**EVM merge** — `evm-rpc` and `evm-ws` consolidated into single `evm` listener. Envoy handles WebSocket upgrade natively. Matches industry standard (Alchemy, Infura, QuickNode all use one URL for HTTP + WS).

**Mode-derived routes** — listeners determined by `NodePortsForMode()`:
- Full/Archive → 4 routes: `evm`, `rpc`, `rest`, `grpc`
- Validator → 0 routes (no public traffic)

**Hostname pattern** — `{deployment}.{protocol}.{domain}`:
```
pacific-1.evm.prod.platform.sei.io   → :8545
pacific-1.rpc.prod.platform.sei.io   → :26657
pacific-1.rest.prod.platform.sei.io  → :1317
pacific-1.grpc.prod.platform.sei.io  → :9090
```

**New platform env var** — `SEI_GATEWAY_DOMAIN` (required) for hostname construction.

## Why

- Deployment names (`rpc`) were conflating with protocol names in subdomains
- Multiple deployments of the same type would collide on hostnames
- EVM HTTP/WS split was non-standard (no EVM provider does this)
- Protocol routing is a Gateway concern, not a per-deployment concern

See design rationale: `SeiKnowledgeBase/memories/node-configuration/networking-design-rationale.md`

## Test plan
- [x] `make test` passes
- [x] `make lint` passes (0 issues)
- [x] `make manifests generate` produces clean CRDs
- [ ] Verify full mode generates 4 HTTPRoutes with correct hostnames
- [ ] Verify validator mode generates 0 HTTPRoutes
- [ ] Verify EVM route covers both HTTP and WebSocket (single hostname)


internal/platform/platformtest/config.go
```diff
@@ -28,5 +28,6 @@ func Config() platform.Config {
 		GenesisRegion:       "us-east-2",
 		GatewayName:         "sei-gateway",
 		GatewayNamespace:    "istio-system",
+		GatewayDomain:       "test.platform.sei.io",
```
bdchatham, line 31: Why do we need this? We shouldn't have test resources which aren't in test file generally.

manifests/samples/seinodedeployment/pacific-1-rpc-group.yaml
```diff
@@ -45,8 +45,7 @@ spec:
     service:
       type: ClusterIP
 
-    gateway:
-      baseDomain: pacific-1.sei.io
+    gateway: {}
```
bdchatham: Do we need the gateway struct? Can we assume it's always configured and it only needs to appear if you are specifying annotations?

internal/platform/platformtest/config.go
bdchatham: The `platformtest` package is the existing shared test fixture pattern — it's used by 6 test files across `node/`, `task/`, and `planner/` packages. Every platform config field (NodepoolName, SnapshotBucket, etc.) lives here. Adding `GatewayDomain` follows the same convention. The package exists specifically because multiple test packages need the same platform config fixture and Go doesn't allow cross-package test helpers without a shared package.

PR #79, opened 2026-04-13
design: InPlace update strategy for SeiNodeDeployment

bdchatham:
## Summary
- LLD for formalizing InPlace as an explicit `UpdateStrategyType` alongside BlueGreen and HardFork
- Sibling design to #67 (HardFork bootstrap simplification), following the same structure and depth
- Designed with input from kubernetes-specialist and platform-engineer agents

## Key Design Decisions
- **No plan/task machinery** — InPlace bypasses the deployment planner entirely, using the existing `ensureSeiNode` in-place propagation path with rollout status tracking layered on top
- **`ensureMarkReady` in `reconcileRunning`** — solves the sidecar restart problem for ALL strategies (pod restarts cause sidecar to block seid indefinitely). This is the highest-priority standalone fix
- **Optional `upgradeHeight` gating** — when set, the controller validates the chain has reached the upgrade height before patching StatefulSets, preventing premature rollouts
- **Simultaneous rollout** — all nodes update at once (chain upgrades are coordinated halts)
- **No auto-rollback** — rolling back a blockchain binary after a chain upgrade leaves the node unable to process new blocks; controller reports Degraded and lets the engineer decide
- **`RolloutStatus` tracking** — per-node convergence status distinct from `DeploymentStatus` (which tracks entrant/incumbent for blue-green)

## Design Covers
- CRD changes (enum, `InPlaceStrategy`, `RolloutStatus` types)
- Controller changes (`detectDeploymentNeeded` branching, rollout status reconciliation)
- Sidecar mark-ready resolution (primary: controller re-submit; future: sidecar self-detect)
- Status reporting (conditions, events, phase transitions)
- Failure modes and recovery
- File-by-file changes and implementation order
- Test plan

📄 Design doc: `.tide/designs/inplace-update-strategy.md`

### DIFF
--- a/.tide/designs/inplace-update-strategy.md
+++ b/.tide/designs/inplace-update-strategy.md
@@ -1,0 +1,415 @@
+# InPlace Update Strategy for SeiNodeDeployment
+
+## Summary
+
+The InPlace update strategy is a lightweight, operator-driven deployment mode that propagates spec changes (image, sidecar, entrypoint) directly to existing SeiNode resources without creating entrant nodes or performing blue-green traffic switching. The controller's role is confined to change propagation, health monitoring, and status reporting. Unlike the nil strategy (no `UpdateStrategy` set), InPlace provides explicit rollout tracking through a `RolloutStatus` on the SeiNodeDeployment, surfacing per-node convergence state so engineers can observe the rollout and detect failures.
+
+The critical technical challenge is the sidecar mark-ready gate: when a pod restarts after an image update, the sidecar starts fresh and returns 503 from `/v0/healthz`, blocking seid startup indefinitely. This design solves that by having the SeiNode controller's Running phase reconciler submit `mark-ready` unconditionally on every reconcile, with a future path for the sidecar to self-resolve on restart.
+
+## Motivation
+
+Today, the controller supports three update paths:
+
+1. **BlueGreen** -- full blue-green with entrant nodes, catch-up wait, traffic switch, and incumbent teardown
+2. **HardFork** -- blue-green variant with halt-height coordination for chain upgrades
+3. **nil (no strategy)** -- `ensureSeiNode` patches child SeiNodes in-place with no orchestration, no status tracking, and no health monitoring
+
+The nil strategy is fire-and-forget. After pushing the change to child SeiNodes, the deployment controller moves on. There is no rollout status, no per-node convergence tracking, and no way for an operator to answer "did the rollout succeed?" without manually inspecting each SeiNode and its pod.
+
+InPlace fills this gap. It is the right strategy when:
+
+- The operator has already waited for the chain to reach upgrade height (or the change is non-disruptive, like a sidecar bump)
+- Creating fresh nodes is unnecessary or wasteful (the existing PVCs hold valid state)
+- The operator wants controller-assisted status reporting without the cost and complexity of blue-green
+
+Formalizing InPlace also makes the nil path clearer: nil means "I don't want any deployment orchestration at all, not even tracking."
+
+## How It Works
+
+### InPlace Rollout Lifecycle
+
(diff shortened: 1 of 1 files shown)


.tide/designs/inplace-update-strategy.md
```diff
@@ -0,0 +1,415 @@
...
+
+The critical technical challenge is the sidecar mark-ready gate: when a pod restarts after an image update, the sidecar starts fresh and returns 503 from `/v0/healthz`, blocking seid startup indefinitely. This design solves that by having the SeiNode controller's Running phase reconciler submit `mark-ready` unconditionally on every reconcile, with a future path for the sidecar to self-resolve on restart.
+
+## Motivation
+
+Today, the controller supports three update paths:
+
+1. **BlueGreen** -- full blue-green with entrant nodes, catch-up wait, traffic switch, and incumbent teardown
+2. **HardFork** -- blue-green variant with halt-height coordination for chain upgrades
+3. **nil (no strategy)** -- `ensureSeiNode` patches child SeiNodes in-place with no orchestration, no status tracking, and no health monitoring
```
bdchatham: I would remove this path going forward. It was a choice to keep the door open and avoid making a decision on this InPlace strategy. Now we are formalizing it with InPlace.

.tide/designs/inplace-update-strategy.md
```diff
@@ -0,0 +1,415 @@
...
+    UpdateStrategyHardFork  UpdateStrategyType = "HardFork"
+    UpdateStrategyInPlace   UpdateStrategyType = "InPlace"
+)
+```
+
+### UpdateStrategy extension
+
+```go
+type UpdateStrategy struct {
+    Type UpdateStrategyType `json:"type"`
```
bdchatham, line 108: Let's make BlueGreen an explicit type.

.tide/designs/inplace-update-strategy.md
```diff
@@ -0,0 +1,415 @@
...
+// InPlaceStrategy configures an in-place rollout.
+type InPlaceStrategy struct {
+    // UpgradeHeight is the block height at which the chain upgrade occurs.
+    // When set, the controller validates via the sidecar that the chain
+    // has reached this height before patching the StatefulSet. This prevents
+    // premature rollouts when manifests are pushed before the chain halts.
+    // When omitted, the update is applied immediately (suitable for
+    // non-upgrade changes like sidecar bumps or config tweaks).
+    // +optional
+    // +kubebuilder:validation:Minimum=1
```
bdchatham: This isn't really possible. The main container has already stopped and the sidecar isn't tracking this progress. For InPlace, I would suggest simplifying to just immediately apply a change in the main container (I would only track the container image changes) and materialize that.

.tide/designs/inplace-update-strategy.md
```diff
@@ -0,0 +1,415 @@
...
+```
+
+### RolloutStatus (new type)
+
+```go
+// RolloutStatus tracks an in-progress in-place rollout. Unlike
+// DeploymentStatus (which tracks entrant/incumbent node sets for
+// blue-green), RolloutStatus tracks per-node convergence of the
+// existing node set.
+type RolloutStatus struct {
```
bdchatham, line 138: Seems like we should combine RolloutStatus and DeploymentStatus. I'd prefer RolloutStatus for the name of the combined data structure.

.tide/designs/inplace-update-strategy.md
```diff
@@ -0,0 +1,415 @@
...
+```go
+func (r *SeiNodeDeploymentReconciler) upgradeHeightReached(ctx context.Context, group *seiv1alpha1.SeiNodeDeployment) bool {
+    strategy := group.Spec.UpdateStrategy
+    if strategy == nil || strategy.InPlace == nil || strategy.InPlace.UpgradeHeight == nil {
+        return true
+    }
+    // Check at least one node has reached the upgrade height
+    // via sidecar status polling (existing sidecar client)
+    // ...
+}
```
bdchatham: I like this idea, but there is an obvious race between when we check this value and when the seid container itself crashes on an upgrade, for example

PR #87, opened 2026-04-15
feat: plan-centric reconcile foundations (Steps 1-4)

bdchatham:
## Summary

Lays the groundwork for making `Plan` the central reconciliation abstraction across all SeiNode phases. This PR contains the **non-behavioral half** — all new code is additive and existing controller behavior is unchanged.

- **CRD**: Add `TargetPhase`/`FailedPhase` to `TaskPlan` so the executor can drive phase transitions atomically with plan completion/failure
- **noderesource package**: Extract StatefulSet, Service, and PVC generators from the controller into `internal/noderesource/`, breaking the import cycle needed for Step 3
- **New tasks**: `ensure-data-pvc`, `apply-statefulset`, `apply-service` — fire-and-forget controller-side tasks registered in the task registry (not yet wired into plans)
- **Executor**: Handles `TargetPhase`/`FailedPhase` on plan terminal states, nils convergence plans (target == current phase), adds `OnPlanComplete`/`OnPlanFailed` callbacks

### What's next (separate PR)

Steps 5-6 will wire these foundations into the actual reconcile loop:
- Phase-aware `BuildPlan` (init plans include infrastructure tasks, Running phase gets convergence plans)
- Mutating `ForNode` that stamps plans directly onto `node.Status.Plan`
- Simplified reconcile loop: `ForNode → ExecutePlan`

## Test plan

- [x] `make test` — all existing tests pass
- [x] `make lint` — zero issues
- [x] `make build` — compiles cleanly
- [x] `make manifests generate` — CRDs regenerated
- [x] Review: executor TargetPhase/FailedPhase handling in `executor.go`
- [x] Review: new task implementations follow existing patterns (compare with `bootstrap_job.go`)
- [x] Review: noderesource extraction preserves all behavior (95.5% coverage)


internal/planner/executor.go
```diff
@@ -29,6 +29,12 @@ type PlanExecutor[T client.Object] interface {
 	ExecutePlan(ctx context.Context, obj T, plan *seiv1alpha1.TaskPlan) (ctrl.Result, error)
 }
 
+// PhaseTransition records a phase change triggered by plan completion or failure.
+type PhaseTransition struct {
+	PrevPhase string
+	NewPhase  string
+}
```
bdchatham: I don't quite understand why we need this?

internal/planner/executor.go
```diff
@@ -37,9 +43,15 @@ type PlanExecutor[T client.Object] interface {
...
+// OnPlanComplete and OnPlanFailed are optional callbacks invoked after the
+// status patch succeeds. They allow the caller to emit metrics or events
+// without coupling the executor to resource-specific concerns.
 type Executor[T client.Object] struct {
-	Client    client.Client
-	ConfigFor func(ctx context.Context, obj T) task.ExecutionConfig
+	Client         client.Client
+	ConfigFor      func(ctx context.Context, obj T) task.ExecutionConfig
+	OnPlanComplete func(ctx context.Context, obj T, transition *PhaseTransition)
+	OnPlanFailed   func(ctx context.Context, obj T, transition *PhaseTransition)
```
bdchatham: Why does the plan need callbacks? Why can't the plan just own this? When the plan is driven, if it hits a case where it is moving to failed or complete state, it must patch the node, for example. Can that not be done? A task and plan are basically just JSON, so I can see where there is a gap. It seems like conditions are the interface we need to use here. If a plan fails, set a generic condition explaining this on so corrective action can be taken or for now, while we have a simple implementation, a human operator can take manual action.

PR #90, opened 2026-04-16
feat: single-patch reconcile model

bdchatham:
## Summary

Both controllers now accumulate all status mutations in-memory and flush with a single `Status().Patch()` at the end of Reconcile. This eliminates the class of optimistic lock conflicts between tasks and the executor, and establishes a clean ownership model.

- **Executor**: Removes `Client` field entirely — pure in-memory mutation engine. Synchronous tasks advance in a loop without intermediate patches. `failTask` is void.
- **Node controller**: Single `statusBase` captured after finalizer. `reconcilePeers` and `observeCurrentImage` return dirty flags. One flush at the end.
- **NodeDeployment controller**: `completePlan`/`failPlan` are now void in-memory mutations. No more re-read after executor patches (because there are none).

### Ownership model

| Layer | Writes to |
|-------|-----------|
| Tasks | Owned resources (StatefulSets, Services, PVCs, sidecar) |
| Executor | In-memory plan state on owning resource |
| Reconciler | Single `Status().Patch()` flush at end |

### What this unblocks

Tasks can now freely mutate the owning resource's status fields (e.g., `status.currentImage`, conditions) because everything is in-memory until the flush. This is the foundation for NodeUpdate plans where tasks need to set conditions and stamp image state.

## Test plan

- [x] `make build` — compiles cleanly
- [x] `make test` — all tests pass (84.6% node controller, 43.5% planner, 32.9% nodedeployment)
- [x] `make lint` — zero issues

### DIFF
--- a/internal/controller/node/controller.go
+++ b/internal/controller/node/controller.go
@@ -70,6 +70,8 @@ func (r *SeiNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return r.handleNodeDeletion(ctx, node)
 	}
 
+	// Finalizer is a metadata Update — must happen before we snapshot
+	// the status patch base because Update changes resourceVersion.
 	if err := r.ensureNodeFinalizer(ctx, node); err != nil {
 		return ctrl.Result{}, err
 	}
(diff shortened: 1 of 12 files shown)


internal/controller/node/controller.go
```diff
@@ -81,38 +83,62 @@ func (r *SeiNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, nil
 	}
 
+	// --- From here, all status mutations are in-memory. ---
```
bdchatham: Let's trim down this comment. Comments like this are a smell. The code should be self-explaining and if it's not, I'd add a concise comment explaining the strategy above the reconcile method.

internal/controller/node/controller.go
```diff
@@ -81,38 +83,62 @@ func (r *SeiNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
-		return result, err
+	// --- Flush all accumulated status mutations in a single patch. ---
+	if statusDirty {
+		if err := r.Status().Patch(ctx, node, statusBase); err != nil {
+			return ctrl.Result{}, fmt.Errorf("flushing status: %w", err)
+		}
+	}
+
+	// Return exec error AFTER the flush so partial progress is persisted.
+	if execErr != nil {
```
bdchatham, line 141: We are swallowing this error if it's non-nil and there is an error on the patch. Let's check for that and log error if that scenario happens so this isn't left burried.

internal/controller/node/peers_test.go
```diff
@@ -39,28 +38,23 @@ func TestReconcilePeers_ResolvesLabelSource(t *testing.T) {
 		Spec: seiv1alpha1.SeiNodeSpec{ChainID: "test-1", Image: "sei:latest", FullNode: &seiv1alpha1.FullNodeSpec{}},
 	}
 
-	r, c := newNodeReconciler(t, node, peer1, peer2)
+	r, _ := newNodeReconciler(t, node, peer1, peer2)
```
bdchatham, line 41: Do we need to return client by this method if it's not used? Is it used in other places?

internal/planner/executor.go
```diff
@@ -101,63 +100,80 @@ func executePlan(
...
+
+		// If the task is not yet complete, stop the loop — the result
+		// carries the appropriate requeue interval.
+		if t.Status != seiv1alpha1.TaskComplete {
+			return result, nil
 		}
-		planActive.WithLabelValues(cn, obj.GetNamespace(), obj.GetName()).Set(0)
-		return ResultRequeueImmediate, nil
+
+		// Task completed synchronously — continue to the next task.
```
bdchatham: Let's remove this comment

internal/planner/executor.go
```diff
@@ -166,41 +182,36 @@ func executePlan(
...
-	kc client.Client,
 	obj client.Object,
 	controller string,
 	plan *seiv1alpha1.TaskPlan,
 	t *seiv1alpha1.PlannedTask,
 	errMsg string,
-) (ctrl.Result, error) {
-	log.FromContext(ctx).Error(fmt.Errorf("task failed: %s", errMsg), "task plan failed", "task", t.Type)
+) {
+	log.Log.Error(fmt.Errorf("task failed: %s", errMsg), "task plan failed", "task", t.Type)
```
bdchatham: Is there a reason to not use FromContext?

Issue #105, opened 2026-04-20
feat: first-class support for importing existing storage into SeiNode


bdchatham:
## Context

Today `ensure-data-pvc` (see `internal/task/ensure_pvc.go`) always tries to `Create()` a fresh PVC for every SeiNode. There's no first-class way to tell a SeiNode: "use this pre-populated volume as your data directory."

The controller currently tolerates pre-existing PVCs *by accident* — `Create()` returns `AlreadyExists`, the task marks Complete and moves on. That's exactly the bug in #104. It works for manual migrations (we just used it to land a 17 TiB pacific-1 archive into k8s from a non-k8s EC2 source) but it's fragile as a general pattern.

This issue tracks the work to promote "import existing storage" to a real spec field, and to fix the underlying create-vs-import conflation at the same time.

## Use cases

1. **AMI/snapshot-based bootstrap.** The team produces weekly AMIs of running pacific-1 archives via a snapshotter pipeline. New k8s archives should bootstrap from the latest snapshot (catching up hours of blocks) instead of resync-from-genesis (months).
2. **One-off migrations.** Moving an existing archive into k8s without re-syncing — already in use, informally.
3. **Orphan adoption.** \`deletionPolicy: Retain\` + the SeiNode finalizer's \`deleteNodeDataPVC\` orphan PVs today. A re-created SeiNode should be able to explicitly adopt an orphan.

## Direction (to be documented in a design PR)

Candidate shapes under evaluation:
- **A** — reference an existing PVC by name (\`spec.dataVolume.import.pvcName\`)
- **D** — reference an existing PV by name (\`spec.dataVolume.import.pvName\`, controller creates the PVC bound to it)
- **C** — reference an EBS snapshot ID (controller calls \`CreateVolume\`)
- **E** — reference a Kubernetes \`VolumeSnapshot\` (k8s-native snapshot path, zero cloud SDK in the controller)

Current lean: **A + D as a single design pass**, **E on the roadmap** for snapshot-restore once that's needed, **C rejected** on cloud-provider-awareness grounds (would force AWS SDK + IRSA into an otherwise cloud-agnostic controller).

## Acceptance (rough — refined in the design PR)

- \`spec.dataVolume.import.{pvcName,pvName}\` fields on SeiNode
- Planner routes to \`validate-imported-pvc\` or \`adopt-pv-and-create-pvc\` task instead of \`ensure-data-pvc\` when import is set
[...]

Issue #104, opened 2026-04-19
ensure-data-pvc marks Complete on AlreadyExists against a terminating PVC (PVC-race bug)


bdchatham:
## Context

Production incident on 2026-04-17 during a `kubectl delete seinode` / recreate cycle on `pacific-1/archive-0-0`: the new SeiNode's `ensure-data-pvc` task reached `Complete`, but the backing PVC was missing a few seconds later, leaving the pod `Pending` with:

```
FailedScheduling: PersistentVolumeClaim "data-archive-0-0" not found
```

### Root cause

The old SeiNode's finalizer (`deleteNodeDataPVC`) was still draining when the new SeiNode's `ensure-data-pvc` task ran. The PVC existed in the API (with `deletionTimestamp` set), so `Create` returned `AlreadyExists`. The task interpreted that as success and marked `Complete`. A moment later the finalizer finished and the PVC disappeared — but the plan had already moved on. The pod was scheduled against a PVC that no longer existed.

This is compounded by the `io2-archive` StorageClass having `reclaimPolicy: Retain` + the SeiNode finalizer explicitly deleting the PVC, but the core bug lives in the task.

## Call site

`internal/task/ensure_pvc.go:45-66` — `ensureDataPVCExecution.Execute`:

```go
if err := e.cfg.KubeClient.Create(ctx, desired); err != nil {
    if apierrors.IsAlreadyExists(err) {
        e.complete()
        return nil
    }
    return fmt.Errorf("creating data PVC: %w", err)
}

e.complete()
return nil
```

The `AlreadyExists` branch:
- Does not re-fetch the PVC to check `DeletionTimestamp`.
- Does not verify the spec matches what we generated (StorageClass, size, access modes).
- Does not wait for `Phase == Bound`.

Any one of those checks would have caught the race. Even the happy-path `Create` success can end with `Complete` before the PVC actually binds (e.g., quota or provisioning delays), but today's incident was specifically the `AlreadyExists`-on-terminating case.

## Reproduction

1. Create a SeiNode with a data volume.
2. `kubectl delete seinode <name>` — finalizer starts deleting the PVC.
3. Recreate the SeiNode quickly (or rely on a parent SeiNodeDeployment which recreates in ~1s).
4. Observe: `ensure-data-pvc` task reports `Complete`, plan proceeds, PVC is missing from the cluster seconds later.

[...]

PR #106, opened 2026-04-20
docs: design for SeiNode import existing storage

bdchatham:
Tracks #105. Related: #104.

## Summary

Direction-setting doc for adding first-class "import existing volume" support to `SeiNode`, so users can stage a pre-populated PVC/PV and reference it from the SeiNode spec — rather than relying on the current `ensure-data-pvc` `Create()`-then-`AlreadyExists` accident (which is #104).

**Decision in the doc**: adopt **Shape A** (\`import.pvcName\`) + **Shape D** (\`import.pvName\`) together. Defer **Shape E** (`VolumeSnapshot` CR) to a later pass once snapshot-restore is directly needed. Reject **Shape C** (EBS snapshot ID with controller-side `CreateVolume`) to avoid making the controller cloud-provider-aware.

## Why this is just the design doc (no code)

This PR is deliberately narrow — prose + tables, no CRD changes, no task implementation. Goal is to get directional sign-off inline via comments, then a follow-up PR with the full LLD and implementation.

## What's left open

Three open questions called out at the bottom of the doc — deletion semantics, validation scope (mutate vs. read-only-fail), init-plan interaction with the bootstrap Job. Each has my current lean, but all three are genuinely open for the LLD.

## Review ask

- Inline comments on specific claims welcome
- Pay special attention to:
  - The "reject Shape C" reasoning (cloud-provider awareness) — if you disagree, the design shifts
  - The three open questions — each needs a decision before LLD work starts
  - The "does NOT cover" list — make sure we're not implicitly promising something we shouldn't

## Test plan

N/A — documentation only. If the direction lands, follow-up PR does the LLD + implementation with unit tests.

### DIFF
--- a/docs/design-seinode-import-volume.md
+++ b/docs/design-seinode-import-volume.md
@@ -1,0 +1,106 @@
+# Design: SeiNode — Import Existing Storage
+
+**Status:** Draft / RFC
+**Date:** 2026-04-20
+**Tracks:** [#105](https://github.com/sei-protocol/sei-k8s-controller/issues/105)
+**Related:** [#104](https://github.com/sei-protocol/sei-k8s-controller/issues/104)
+
+## Problem
+
+Today `internal/task/ensure_pvc.go` always calls `Create()` for a SeiNode's data PVC. There is no first-class way to express "use this pre-populated volume." The controller *accidentally* tolerates pre-existing PVCs — `Create()` returns `AlreadyExists`, the task marks Complete, and reconciliation proceeds. This is the bug in #104.
+
+Three use cases want this behavior to be a real feature, not an accident:
+
+1. **AMI/snapshot-based bootstrap.** Weekly AMIs of running pacific-1 archives exist; new k8s archives should bootstrap from the latest snapshot (catching up hours of blocks) instead of resync-from-genesis (months).
+2. **One-off migrations.** Moving an existing archive into k8s without re-syncing — currently done by staging a PVC with the expected name and relying on the accidental path.
+3. **Orphan adoption.** `deletionPolicy: Retain` + the SeiNode finalizer's `deleteNodeDataPVC` orphan PVs. A re-created SeiNode should be able to explicitly adopt an orphan.
+
+The feature and the bug share a refactor: splitting `ensure-data-pvc` into a create-path and an import-path forces us to write the validation logic that #104's `AlreadyExists` swallow is currently hiding.
+
+## Shapes evaluated
+
+| Axis | **A** — import PVC by name | **B** — import EBS volume ID | **C** — import EBS snapshot ID |
+|---|---|---|---|
+| User ergonomics | Two-step (pre-create PV+PVC, then reference). Familiar to k8s operators. | One-step from user POV. Controller owns PV. | One-step; covers weekly-AMI use case end-to-end. |
+| Controller complexity | Near-zero — conditional skip of `ensure-data-pvc`, plus a validating read. | Moderate — controller synthesizes a static PV with `csi.volumeHandle`, sets `claimRef`. No cloud SDK. | High — AWS SDK, IRSA creds, region plumbing, async `CreateVolume` polling, idempotency, AZ selection, backoff/retry. |
+| Cloud-provider awareness | None | Thin — names a volume handle but never calls AWS | Crosses into cloud-provider controller territory |
+| Failure modes | PVC missing / wrong size / wrong AZ / already bound elsewhere — all observable via k8s API | Volume in wrong AZ, attached elsewhere, wrong fsType — observable when CSI fails to attach | All of B, plus: snapshot doesn't exist, quota exceeded, AWS throttling, half-created volumes on controller crash, orphan-volume billing leaks |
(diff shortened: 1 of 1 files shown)


docs/design-seinode-import-volume.md
```diff
@@ -0,0 +1,106 @@
...
+      # Optional:
+      reclaimPolicy: Preserve        # override the PV's default reclaim behavior on SeiNode deletion
+```
+
+Planner behavior: if `spec.dataVolume.import` is set, `ensure-data-pvc` is replaced with a new task — `validate-imported-pvc` (Shape A) or `adopt-pv-and-create-pvc` (Shape D). Successor tasks (`apply-statefulset`, `apply-service`, etc.) are unchanged. Snapshot-restore and bootstrap-Job tasks are skipped when importing — the data is by definition already present.
+
+## Open questions (for the LLD)
+
+1. **Deletion semantics for imported volumes.** When a SeiNode with `import.pvcName` is deleted, does `deleteNodeDataPVC` still delete the PVC?
+   - **(a)** Always preserve imported PVCs (safe, but deletes don't fully clean up)
```
bdchatham: I like this approach.

docs/design-seinode-import-volume.md
```diff
@@ -0,0 +1,106 @@
...
+
+1. **AMI/snapshot-based bootstrap.** Weekly AMIs of running pacific-1 archives exist; new k8s archives should bootstrap from the latest snapshot (catching up hours of blocks) instead of resync-from-genesis (months).
+2. **One-off migrations.** Moving an existing archive into k8s without re-syncing — currently done by staging a PVC with the expected name and relying on the accidental path.
+3. **Orphan adoption.** `deletionPolicy: Retain` + the SeiNode finalizer's `deleteNodeDataPVC` orphan PVs. A re-created SeiNode should be able to explicitly adopt an orphan.
+
+The feature and the bug share a refactor: splitting `ensure-data-pvc` into a create-path and an import-path forces us to write the validation logic that #104's `AlreadyExists` swallow is currently hiding.
+
+## Shapes evaluated
+
+| Axis | **A** — import PVC by name | **B** — import EBS volume ID | **C** — import EBS snapshot ID |
```
bdchatham, line 22: I prefer Option A

docs/design-seinode-import-volume.md
```diff
@@ -0,0 +1,106 @@
...
+
+1. **Deletion semantics for imported volumes.** When a SeiNode with `import.pvcName` is deleted, does `deleteNodeDataPVC` still delete the PVC?
+   - **(a)** Always preserve imported PVCs (safe, but deletes don't fully clean up)
+   - **(b)** Respect the PV's `persistentVolumeReclaimPolicy` (Retain vs Delete) — most k8s-idiomatic
+   - **(c)** Add explicit `spec.dataVolume.import.reclaimPolicy` that overrides
+   
+   Leaning (b) with (c) as override.
+
+2. **What does "validate" actually check, and do we ever mutate?** Size is obvious (PV capacity ≥ requested). But:
+   - Do we require PVC labels to match `ResourceLabels`?
```
bdchatham: Yes, we need to have a spec/requirements for exactly what the pvc needs.

docs/design-seinode-import-volume.md
bdchatham: Done — Shape A is now the only adopted shape in the Decision section. Shape D is moved to 'Additional options considered' as evaluated-but-not-adopted, with a note that orphan adoption is reachable via Shape A (operator creates a PVC bound to the orphan PV, then references it). Less CRD surface, less task-path branching. Pushed in 3e62e55.

docs/design-seinode-import-volume.md
bdchatham: Agreed, flipped the lean. Moved from open-question to a decided section — the controller will NOT delete imported PVCs on SeiNode deletion (deleteNodeDataPVC is a no-op for imported volumes). Rationale: operators who opted into import are managing storage lifecycle externally; controller shouldn't touch storage it didn't create. Pushed in 3e62e55.

docs/design-seinode-import-volume.md
bdchatham: Added a new 'Requirements for an imported PVC' section with a seven-requirement table (exists, not terminating, Bound, RWO, capacity, PV integrity). Also spelled out what the controller explicitly does NOT check (fstype, content, labels, SC) with the reasoning. No longer an open question — it's a decided contract. Pushed in 3e62e55.

bdchatham:
Addressed your latest feedback in c86790b:

- Removed the separate task (`validate-imported-pvc`) idea entirely
- Import is now strictly a PVC-source substitution inside `ensure-data-pvc`
- Init plan is unchanged; every successor task runs as-is
- Explicit section saying the controller does NOT skip any steps, does NOT refuse to run `configure-genesis` on imported data, does NOT take responsibility for data contents
- Operator is trusted to provide a PVC that works downstream; failures surface through the normal plan-Failed channel

Only one open question remains for the LLD (retry budget semantics).

bdchatham:
Pushed 98a1179 — retry semantics decided: retry indefinitely with exponential backoff, surface via Condition. All direction-level open questions are now resolved. LLD work is implementation specifics (backoff curve, Condition schema, tests). Ready for merge whenever you're happy with the direction.

PR #108, opened 2026-04-21
feat: decouple snapshot generation from publishing

bdchatham:
## Summary

Implements the merged design in `docs/design-snapshot-publish-config.md` (#107). Operators can now generate local Tendermint snapshots without also uploading them to S3, and the snapshot-config surface is shaped so future snapshot flavors slot in without renaming fields.

- Refactored `SnapshotGenerationConfig` to nest a `TendermintSnapshotGenerationConfig` sub-struct; added optional `Publish` struct whose *presence* enables S3 upload. Matches the existing `StateSyncSource struct{}` / `SnapshotSource` presence-as-toggle pattern already used in `common_types.go`.
- Re-wired the existing seictl `TaskTypeSnapshotUpload` sidecar task as a fire-and-forget plan task. Inserted into the init progression immediately before `mark-ready` **only** when `publish` is set. Restores controller-side plumbing that was deleted in #89 with the monitor-task subsystem, now as a plan task.
- Cross-field validation (empty `snapshotGeneration`, `publish + keepRecent < 2`) lives in each planner's `Validate(node)` hook with its own mode-prefixed error (`fullNode:` / `archive:`). No admission webhook introduced; no shared helper takes a mode parameter.
- Samples (`pacific-1-snapshotter.yaml`, `pacific-1-state-syncer.yaml`) updated to the new shape, preserving their prior publishing behavior.
- CRDs regenerated via `make manifests generate`; DeepCopy regenerated. No hand edits to generated files.

## Shape of the new surface

```yaml
# Generate 5 local snapshots + upload to S3 (previous default behavior)
snapshotGeneration:
  tendermint:
    keepRecent: 5
    publish: {}

# Generate 3 local snapshots, do NOT upload (new capability)
snapshotGeneration:
  tendermint:
    keepRecent: 3

# No snapshot generation
# (field omitted)
```

## Verification

- `make build` — succeeds
- `make lint` — 0 issues
- `make test` — all packages pass
- Independent cross-review of the diff vs. the design doc: all 14 checks COMPATIBLE (type shape, planner overrides, validation shape & rules, plan-task wiring on both non-bootstrap and post-bootstrap paths, exclusion from the bootstrap Job's own sidecar, samples, tests, generated files)

## Test plan

- [ ] Apply an updated sample SeiNode with `publish: {}` and observe `snapshot-upload` appearing in `.status.plan` between `config-validate` and `mark-ready`
- [ ] Apply the same sample with `publish` omitted and confirm `snapshot-upload` is NOT in the plan; confirm seid still writes snapshots to the data volume at the configured interval
- [ ] Apply a spec with `publish: {}` and `keepRecent: 1` — expect the planner's `Validate()` to reject with a clear error mentioning the upload algorithm
- [ ] Apply `snapshotGeneration: {}` (empty) — expect the planner's `Validate()` to reject as a typo-shape
- [ ] Spot-check one existing snapshotter node: delete + recreate under the new shape; confirm the sidecar begins uploading after `mark-ready`
- [ ] Review generated CRD YAML in `manifests/sei.io_seinodes.yaml` and `manifests/sei.io_seinodedeployments.yaml` for the new nested shape


internal/planner/bootstrap.go
```diff
@@ -105,13 +108,26 @@ func buildBootstrapPlan(
...
-	return prog
+	return maybeInsertSnapshotUpload(prog, node)
+}
+
+// maybeInsertSnapshotUpload inserts the snapshot-upload sidecar task into prog
+// immediately before mark-ready when the node is configured to publish
+// Tendermint snapshots. Placement after config-apply/config-validate and before
+// mark-ready means the sidecar's upload loop is armed before seid starts
+// producing snapshots. Returns prog unchanged when publish is not set.
+func maybeInsertSnapshotUpload(prog []string, node *seiv1alpha1.SeiNode) ([]string, error) {
```
bdchatham: I'd like to have a better name than "maybe" in the method. I'd suggest just having this code directly in the caller method. It's a little unwieldy as is.

internal/planner/full.go
```diff
@@ -43,8 +46,8 @@ func (p *fullNodePlanner) BuildPlan(node *seiv1alpha1.SeiNode) (*seiv1alpha1.Tas
 
 func (p *fullNodePlanner) controllerOverrides(node *seiv1alpha1.SeiNode) map[string]string {
 	sg := node.Spec.FullNode.SnapshotGeneration
-	if sg == nil {
+	if sg == nil || sg.Tendermint == nil {
 		return nil
 	}
-	return seiconfig.SnapshotGenerationOverrides(sg.KeepRecent)
+	return seiconfig.SnapshotGenerationOverrides(sg.Tendermint.KeepRecent)
```
bdchatham, line 52: Honestly, I think it'd be good to rename this to TendermintSnapshotGenerationOverrides but not a requirement for this PR. Would be nice to create a short issue on the sei-config repo for this.

internal/planner/planner.go
```diff
@@ -320,6 +322,25 @@ func SnapshotGeneration(node *seiv1alpha1.SeiNode) *seiv1alpha1.SnapshotGenerati
 	}
 }
 
+// validateSnapshotGeneration enforces cross-field rules on SnapshotGenerationConfig
+// that the OpenAPI schema cannot express: empty config is a typo, and
+// tendermint.keepRecent must be >= 2 when publish is set. The returned error
+// carries no mode-specific prefix — callers wrap with their mode name (e.g.,
+// fmt.Errorf("fullNode: %w", err)) so the mode label lives next to the
+// mode-specific Validate method.
```
bdchatham: Overly verbose comment.

internal/planner/bootstrap.go
bdchatham: Good call — inlined in both callers and deleted the helper. Pushed in a9d9d12.

internal/planner/full.go
bdchatham: Agreed on the rename. Filed tracking issue on sei-config: sei-protocol/sei-config#6. Keeping this PR scoped to controller-side changes only.

internal/planner/planner.go
bdchatham: Trimmed to one line. Pushed in a9d9d12.

PR #112, opened 2026-04-21
docs: LLD for SeiNode import-existing-storage feature

### LINKED ISSUE #104: ensure-data-pvc marks Complete on AlreadyExists against a terminating PVC (PVC-race bug)

bdchatham:
Tracks #105. Direction doc already merged: \`docs/design-seinode-import-volume.md\`.

## Summary

Companion LLD to the merged direction doc. Scope is strictly bounded to Shape A (import PVC by name) plus the #104 create-path fix — no new shapes, fields, or use cases. This is a design-review PR; no code changes.

## What's in the LLD

1. **CRD schema** — \`spec.dataVolume.import.pvcName\` field, optional, immutable, DNS1123-validated
2. **Task refactor** — \`ensure-data-pvc\` branches on \`import.pvcName\`:
   - Create path: \`Get\`-then-\`Create\`, fails if a non-controlled PVC already exists (fixes #104)
   - Import path: validates the named PVC against the seven requirements from the direction doc
3. **Condition schema** — new \`ImportPVCReady\` Condition with a fixed Reason vocabulary (treated as a public alerting contract)
4. **Backoff curve** — 5s → 10s → 30s → 1m → 5m → 15m cap, tracked statelessly via \`PlannedTask.SubmittedAt\` age
5. **Finalizer** — one-line skip: don't delete imported PVCs on SeiNode deletion
6. **Condition ownership** — planner owns it (matches existing \`NodeUpdateInProgress\` pattern); task writes Reason into \`PlannedTask.Error\` for the planner to read
7. **Tests** — per-requirement table-driven unit tests + integration-style test for the happy-path and eventually-becomes-valid-later path
8. **Rollout** — no migration needed; additive spec field

## Where reviewer attention would be most useful

The LLD makes several interpretation calls on top of the direction doc. Most likely to draw pushback, in rough order:

### 1. Terminal vs. transient classification of validation failures
The direction doc says "retry indefinitely." But some failures (wrong access modes, too-small capacity) have no spontaneous recovery — the operator must delete/recreate the PVC. I split the failure modes by recoverability: transient ones loop indefinitely; structurally-broken ones mark the plan Failed. **Alternative strict reading**: all failures are transient, always loop. Need your call on §3 / §4.

### 2. Reason strings as a "public contract" for alerting
I committed to ten specific CamelCase Reason strings and framed them as stable for Prometheus alerting. If the team doesn't want to commit to stable alerting keys this early, that framing can go — but then we'd version the Reasons freely. **Your call on §4.**

### 3. Condition type name: \`ImportPVCReady\`
Alternatives: \`DataVolumeImported\`, \`ImportedPVCBound\`, \`ImportPVCValid\`. Picked \`ImportPVCReady\` for verb-noun parity with \`PodReady\`/\`NodeReady\`. §4.

### 4. Exact backoff curve (5s–15m cap)
Other defensible curves exist (e.g., 1s start; 30min cap). Justified in §3 but the numbers are a judgment call. The stateless-via-SubmittedAt-age implementation is also coarser than a persisted attempt counter.

### 5. Immutability of \`import.pvcName\`
Strict: forces delete/recreate to re-point. Easier to reason about because the StatefulSet is pinned to one PVC name anyway, but real ergonomics tax. §1.

### 6. Crash-recovery exception in the create-path
The #104 fix says "if a PVC with the expected name already exists when \`import\` is NOT set, treat as an error." I added a \`metav1.IsControlledBy\` exception so we don't error out if *we* created the PVC on a previous reconcile but didn't flush status. This softens the strict reading. §2.

### 7. Task mutating \`PlannedTask.Error\` to propagate Reasons to the planner
Reuses an existing field instead of adding a new Condition-propagation channel. Cleaner but a narrow break of the "tasks don't touch plan state" separation. §6.

### 8. \`persistentvolumes:get;list;watch\` RBAC addition
Minor, but cluster-wide (PVs aren't namespaced). Needed for requirement #6/7 (PV integrity checks). §2.

## Implementation plan (post-merge)

This PR is design-only. Implementation follows as one or more PRs:
- CRD changes + regen (\`make manifests generate\`)
- Task refactor + tests
- Finalizer skip
- Planner Condition plumbing
- RBAC + Helm/Flux deploy

No urgency. The direction doc + this LLD stand as decision records until someone picks up implementation.

## Test plan for this PR

N/A — documentation only.

### DIFF
--- a/docs/design-seinode-import-volume-lld.md
+++ b/docs/design-seinode-import-volume-lld.md
@@ -1,0 +1,496 @@
+# Design: SeiNode — Import Existing Storage (LLD)
+
+**Status:** Draft / LLD
+**Date:** 2026-04-21
+**Tracks:** [#105](https://github.com/sei-protocol/sei-k8s-controller/issues/105)
+**Related:** [#104](https://github.com/sei-protocol/sei-k8s-controller/issues/104), [`docs/design-seinode-import-volume.md`](design-seinode-import-volume.md)
+
+This is the companion LLD to the direction doc. The direction doc fixes *what* and *why*; this doc fixes *how*. Scope is strictly bounded to Shape A (import PVC by name) plus the #104 create-path fix. No new shapes, no new fields, no new use cases.
+
+## 1. CRD schema changes
+
+A single new optional sub-struct is added to `SeiNodeSpec` in `api/v1alpha1/seinode_types.go`. Field naming follows the spec sketch in the direction doc (`spec.dataVolume.import.pvcName`) and k8s API conventions (lowerCamelCase JSON tags, PascalCase Go fields, no acronyms beyond `PVC`).
+
+```go
+// SeiNodeSpec additions (api/v1alpha1/seinode_types.go)
+
+// DataVolume configures the data PersistentVolumeClaim for this node.
+// When omitted, the controller creates a PVC using the node's mode-default
+// storage class and size (see noderesource.DefaultStorageForMode).
+// +optional
+DataVolume *DataVolumeSpec `json:"dataVolume,omitempty"`
+
+// DataVolumeSpec configures how the data PVC is sourced.
+type DataVolumeSpec struct {
+    // Import references a pre-existing PersistentVolumeClaim in the same
+    // namespace as the SeiNode, instead of creating a new one. The
+    // controller validates the referenced PVC but never mutates it.
+    //
+    // When Import is set, the controller never deletes the referenced PVC
+    // on SeiNode deletion — storage lifecycle is the operator's responsibility.
+    // +optional
+    Import *DataVolumeImport `json:"import,omitempty"`
+}
+
+// DataVolumeImport names a pre-existing PVC to adopt as this node's data volume.
+type DataVolumeImport struct {
+    // PVCName is the name of a PersistentVolumeClaim in the SeiNode's
+    // namespace. The PVC must be Bound, ReadWriteOnce, and sized at or above
+    // the node mode's default storage size. Immutable after creation.
+    //
+    // +kubebuilder:validation:MinLength=1
+    // +kubebuilder:validation:MaxLength=253
+    // +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`
+    // +kubebuilder:validation:XValidation:rule="self == oldSelf",message="pvcName is immutable"
+    PVCName string `json:"pvcName"`
+}
+```
+
+### Why these markers
+
+- Length + pattern mirror the k8s DNS1123Label constraint on PVC names — catches typos at admission time.
+- `XValidation: self == oldSelf` makes the reference immutable. Swapping an imported PVC out from under a running SeiNode has no defined semantics; force delete-and-recreate if the operator wants to re-point.
+- `DataVolume` as pointer keeps `omitempty` clean so existing SeiNodes serialize identically to today.
+- No default `storageClassName`/`size` sub-fields — Shape A is "name a PVC, nothing else" per the direction doc.
+
+### Regenerated artifacts
+
+- `zz_generated.deepcopy.go` — new `DeepCopy`/`DeepCopyInto` methods for `DataVolumeSpec` and `DataVolumeImport`; updated `SeiNodeSpec.DeepCopyInto` to copy the pointer. Produced by `make generate`.
+- `manifests/sei.io_seinodes.yaml` and `config/crd/sei.io_seinodes.yaml` — new `spec.dataVolume.import.pvcName` subtree with validation constraints. Produced by `make manifests`.
+
+### Backward compatibility
+
+Existing SeiNodes serialize `SeiNodeSpec` without the `dataVolume` key. The new CRD schema makes `dataVolume` optional, so old objects remain valid. The controller reads `node.Spec.DataVolume` as nil and takes the create path unchanged.
+
+### Spec-unset vs. spec-empty
+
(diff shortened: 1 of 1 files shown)


docs/design-seinode-import-volume-lld.md
```diff
@@ -0,0 +1,496 @@
...
+// +kubebuilder:rbac:groups="",resources=persistentvolumes,verbs=get;list;watch
+```
+
+This is the only RBAC addition. `make manifests` regenerates `manifests/role.yaml`.
+
+### Backoff curve
+
+Per attempt → minimum wall time since `SubmittedAt` before we make another Get:
+
+| Attempt | Min elapsed | Cumulative wall time |
```
bdchatham: My only concern in this entire design is if this backoff policy is overkill and injects unnecessary complexity into the task and how we track it in its opaque json form on the resource. 

Other than that, design is really great.

bdchatham: Fair, simplified in c44be73. Dropped the custom backoff entirely — task now just validates on every reconcile at the executor's existing 5s TaskPollInterval. No importBackoffSchedule, no deriveAttempt, no backoffElapsed, no wall-time arithmetic hidden behind SubmittedAt. The visible state on the resource is just SubmittedAt + Error + ImportPVCReady Condition.

Load analysis: 720 Gets/hr per stuck import hits the controller-runtime informer cache (local-memory reads, not API-server round trips), so the 'stretch the interval to reduce load' argument was doing no real work. Terminal-failure path unchanged — unrecoverable validation failures still fail the plan immediately.

PR #113, opened 2026-04-21
feat: SeiNode import existing PVC + ensure-data-pvc create-path fix

### LINKED ISSUE #104: ensure-data-pvc marks Complete on AlreadyExists against a terminating PVC (PVC-race bug)

### LINKED ISSUE #105: feat: first-class support for importing existing storage into SeiNode

bdchatham:
Closes #105. Fixes #104 as part of the create-path refactor.

Implements the merged design:
- Direction: \`docs/design-seinode-import-volume.md\` ([PR #106](https://github.com/sei-protocol/sei-k8s-controller/pull/106))
- LLD: \`docs/design-seinode-import-volume-lld.md\` ([PR #112](https://github.com/sei-protocol/sei-k8s-controller/pull/112))

## Summary

New optional spec field \`spec.dataVolume.import.pvcName\` lets a SeiNode adopt a pre-existing PVC rather than provisioning one. The controller validates the referenced PVC against the seven requirements from the LLD and never mutates it. Transient failures loop at the executor's existing poll interval; unrecoverable failures mark the plan Failed with a specific \`Reason\`. Status surfaced via a new \`ImportPVCReady\` Condition on SeiNode.

The \`ensure-data-pvc\` create path is refactored to Get-then-Create, treating an existing non-controlled PVC as a terminal error rather than silently binding to it (fixes #104).

## What changed

**CRD / API types** (\`api/v1alpha1/seinode_types.go\`, +generated)
- New \`DataVolumeSpec\` with optional \`Import *DataVolumeImport\`
- \`DataVolumeImport.PVCName\` — immutable, DNS1123-validated, 1-253 chars
- \`ConditionImportPVCReady\` constant
- Ten \`ReasonImport*\` constants documented as a public alerting contract

**Task** (\`internal/task/ensure_pvc.go\`)
- \`Execute()\` branches on presence of \`import.pvcName\`
- \`executeCreate\` — Get-then-Create, ownership-checked. Existing PVC controlled by this SeiNode = Complete (crash recovery). Existing PVC NOT controlled = Terminal (#104 fix).
- \`executeImport\` — seven-check validation with transient/terminal classification per LLD §3
- Transient failures write \`"<Reason>: <Message>"\` into \`PlannedTask.Error\` for the planner to read
- Uses executor's existing \`TaskPollInterval\` — no custom backoff

**Planner** (\`internal/planner/planner.go\`)
- New \`ReconcileImportPVCCondition\` mirroring the existing \`NodeUpdateInProgress\` pattern
- Sets/removes \`ImportPVCReady\` based on current task state
- Called from the reconciler between plan execution and status flush

**Finalizer** (\`internal/controller/node/controller.go\`)
- \`deleteNodeDataPVC\` skips deletion when \`import.pvcName\` is set
- Operator owns imported storage lifecycle

**RBAC**
- \`persistentvolumes: get/list/watch\` — required for PV consistency checks (requirements 6-7). No write verbs.

## Tests

- **18 unit tests** in \`internal/task/ensure_pvc_test.go\`: create-path happy path, create-path crash-recovery idempotence, create-path non-controlled-conflict (#104 guard), import happy path, each of seven validation requirements in transient + terminal classification, late-PVC convergence, poll-cadence regression guard.
- **5 integration tests** in \`internal/controller/node/import_pvc_test.go\`:
  - \`TestController_Import_Happy_ReachesRunning\`
  - \`TestController_Import_LatePVC_Converges\`
  - \`TestController_Import_TerminalFailure_MarksFailed\`
  - \`TestController_Import_Deletion_PreservesPVC\`
  - \`TestController_NoImport_Deletion_DeletesPVC\` (regression guard)
- All existing tests continue to pass.

## Deliberately deferred

- **Metrics and Events from LLD §8** — three counters (\`importPVCValidationTotal\`, \`importPVCTerminalTotal\`, \`importPVCTimeToValid\`) and two Events (\`PVCImportValidated\`, \`PVCImportFailed\`). The feature is functionally complete without them. Keeping this PR focused on correctness + refactor; observability lands in a small follow-up PR once we've shipped this and can instrument real usage.

## Deviations from the LLD (flagged)

Two small implementation decisions worth naming:

1. **\`statusDirty\` flip for condition-only mutations**. The reconciler's existing \`statusDirty\` flag gates the status flush. To ensure a condition-only change gets persisted, added a 3-line compare helper (\`conditionsSnapshot\` + \`conditionsEqual\`) that flips \`statusDirty\` when the condition list mutated. This is implementation plumbing not called out in the LLD but follows directly from "condition is set before flush — must trigger flush when it changes."

2. **Transient API errors on Get**. When \`Get\` returns an unexpected non-\`NotFound\` error (networking blip), classified as transient with \`PVCNotFound\` / \`UnderlyingPVMissing\` reason. The reason is slightly imprecise but behavior is correct (retry). Alternative would be to return the error and let the executor retry at a higher layer. Easy to flip if reviewers prefer.

## Test plan

- [x] \`make test\` — all pass (existing + 23 new)
- [x] \`make lint\` — \`0 issues\`
- [x] \`make generate manifests\` — everything regenerated, committed
- [x] No unintended file modifications
- [x] Post-merge: deploy to a dev cluster, verify archive SeiNode with \`import.pvcName\` referencing a pre-populated PVC reaches Running
- [x] Post-merge: verify existing archives without the field continue to behave exactly as before (backward compat)

## Manual migration context

This PR is what would let us do today's pacific-1 archive import as a first-class feature rather than as a workaround. The manual trick we used yesterday (pre-create PVC with matching name, rely on the \`AlreadyExists\` bug) becomes \`spec.dataVolume.import.pvcName: data-archive-0-0\` and the controller handles it properly.


bdchatham:
Force-pushed a cleanup rebase addressing both specialists' review findings. Summary of changes against previous HEAD (897d2d6 → 72fca2c):

**Structural fix (load-bearing):** Removed the `recordTransient` / `PlannedTask.Error` message-channel idiom. The task now calls `setImportPVCCondition(node, ...)` inline in `executeImport`, which mutates `node.Status.Conditions` directly. Validation returns `error` with `Terminal(...)` wrapping used for the terminal class — matches the existing convention already used elsewhere.

**Planner invariant preserved:** `internal/planner/planner.go` has **zero lines changed** — deleted `ReconcileImportPVCCondition`, `findTaskByType`, `parseTaskError`, `setImportPVCCondition` and all supporting plumbing (~100 lines). The condition projection now lives where the data lives: in the task.

**Reconciler:** Reverted `conditionsSnapshot` / `conditionsEqual` per-call `statusDirty` machinery. Relying on main's `apiequality.Semantic.DeepEqual(before.Status, node.Status)` (from #129) which already covers condition changes for free. Controller diff is now **9 lines**: one RBAC marker + one finalizer-skip block.

**Reasons collapsed 10 → 3:**
- `ReasonPVCValidated` (success)
- `ReasonPVCNotReady` (transient — keep polling)
- `ReasonPVCInvalid` (terminal — fail plan)

Specific failure cause goes in the condition `Message`, not `Reason`. Deleted the "public alerting contract" docstring.

**Storage-class check dropped:** imports are an explicit operator act; matching PVC `StorageClassName` to workload needs is the importer's responsibility. Documented inline in `executeImport` and on `DataVolumeSpec`.

**Immutability tightened:** added CEL rule on `DataVolumeSpec` itself — `(!has(oldSelf.import) || has(self.import))` — so a user can't clear the whole `spec.dataVolume.import` sub-struct to bypass the `pvcName` immutability rule and unintentionally leave the controller to delete externally-managed storage on finalizer.

**Lean pass:** Inlined `importState` enum, `importPVCName` helper, `containsAccessMode` wrapper. Deleted breadcrumb comments (kept `// Fixes #104` and the SC-responsibility note — both carry non-obvious WHY).

**Rebase:** Now on top of main. No longer deletes #129's sidecar-probe / mark-ready-reapply work.

**Stats:**
- Before: 1284 insertions, 5 deletions, 13 files (against previous main)
- After: 1085 insertions, 8 deletions, 12 files (against current main)

**Tests:** all 18 unit + 5 integration pass with the revised assertions. `make lint` → 0 issues. CRDs/deepcopy regenerated.

PR #129, opened 2026-04-23
feat: sidecar mark-ready re-apply on pod replacement

### LINKED ISSUE #127: sei-sidecar mark-ready is not re-applied on pod replacement — stuck validators require manual intervention
## Symptom

When a seid pod is replaced (by kubelet after chaos, node drain, OOM, crash, or any kind of reschedule), the new pod comes up with containers running but stuck:

- `sei-sidecar`: Readiness probe failing with HTTP 503.
- `seid`: Startup probe failing with HTTP 503.

Pod stays in this state indefinitely. No block production by that validator. In dev release-6-5 we observed this for 2+ minutes on a pod targeted by chaos-mesh PodChaos (`action: pod-kill`) — would have stayed stuck forever without intervention.

## Workaround (current)

Port-forward into the sidecar port (default 7777) and POST the mark-ready task:

```
kubectl -n <ns> port-forward <pod> 17777:7777 &
curl -sS -X POST http://localhost:17777/v0/tasks \
  -H "Content-Type: application/json" \
  -d '{"type":"mark-ready"}'
```

The sidecar accepts the task and Grafana shows the validator rejoining consensus within ~20s.

## Why this happens

On initial SND creation, the controller (presumably as part of its task plan) sends mark-ready to each new pod's sidecar after it reaches Running. But on pod *replacement* — same pod name, new container — the controller doesn't notice it needs to re-send mark-ready. The sidecar's `/healthz` stays at 503 until it sees a mark-ready, and seid's startup probe is chained behind the sidecar's readiness. Net result: pod is alive but not actually serving.

## Proposed fix (per discussion with @bdchatham)

Add a **proactive status check on reconciliation** — on every reconcile, for each owned pod, check whether it has been marked ready. If not (e.g., sidecar `/healthz` returns 503 while pod is Running), issue the mark-ready task. Idempotent: sidecar already has deduplication on task ID, so re-sending is safe.

This converges the "sidecar needs mark-ready" state into the regular reconcile loop instead of assuming it's a one-shot handled at creation.

## Test that should pass after fix

1. Create an SND, wait for Ready.
2. `kubectl delete pod <seid-pod>` — kubelet recreates via StatefulSet.
3. Within ~30s: the new pod should be Ready, validator rejoins consensus, block production by this validator resumes. No manual intervention.

## Discovered via

Chaos test 04 (pod-kill) on dev release-6-5. See platform repo `clusters/dev/release-6-5/results/` for the run log.

## Related

- Prior instance: gprusak soak on prod hit the same pattern on rescheduled pods (platform repo, internal investigation — didn't file at the time because the workaround unblocked us)

bdchatham:
Closes #127.
Implements design #128.

## Summary

Proactively detects a stale sidecar mark-ready gate on already-Running SeiNodes and re-issues `mark-ready` via the existing plan system. Pod replacements (chaos, OOM, node drain, kubelet reschedule) no longer leave validators stuck requiring manual `kubectl port-forward` + POST intervention.

## Shape after three review iterations

**`planner.NodeResolver`** — a struct in the planner package that holds the sidecar-client factory. `ResolvePlan` is a method on it. The reconciler is sidecar-agnostic: it wires the factory once in `cmd/main.go` and calls `r.Planner.ResolvePlan(ctx, node)`. No sidecar types in reconciler method signatures.

**Drift detection** lives in `buildRunningPlan`:
- Image drift wins (its plan ends with `MarkReady`, resolving the sidecar in the same plan).
- Otherwise, if `SidecarReady == False/NotReady`, return a one-task `MarkReady` plan.
- `SidecarReady == Unknown` (probe unreachable or errored) does not trigger a plan — re-probe next tick.

**Probe behavior** (in `planner/sidecar_probe.go`):
| Healthz | Condition | Action |
|---|---|---|
| HTTP 200 | `True/Ready` | no plan |
| HTTP 503 | `False/NotReady` | one-task `MarkReady` plan |
| network error | `Unknown/Unreachable` | no plan — re-probe next tick |

Probe runs only when `Phase == Running` (init plans own the sidecar during Initializing). `ResolvePlan` short-circuit on an already-Active plan prevents spawning a new `MarkReady` mid-flight.

**Reconciler simplification**: scattered `statusDirty = true` flags replaced with a single diff-at-end. Status is patched iff `!apiequality.Semantic.DeepEqual(before.Status, node.Status)`, using `MergeFromWithOptimisticLock` to preserve resourceVersion precondition against executor-side patches. Any future drift check that mutates status is automatically covered.

## Surface area

- **New condition:** `ConditionSidecarReady` on `SeiNode`.
- **New metrics:**
  - `sei.controller.seinode.sidecar_health_probes` (counter, labels: `controller`, `namespace`, `outcome ∈ {ready, not_ready, unreachable}`).
  - `sei.controller.seinode.sidecar_ready` (observable gauge, labels: `namespace`, `name`; `1=ready, 0=otherwise`). Lets PrometheusRule alerts fire without kube-state-metrics needing `CustomResourceStateMetrics` configured for our CRD.
- **New events:** `SidecarReadinessLost` (Warning), `SidecarReadinessRestored` (Normal).
- **Interface widening:** `task.SidecarClient` gains `Healthz(ctx) (bool, error)`. Seictl's concrete `*sidecar.SidecarClient` already implements this.
- **Reconciler field:** `SeiNodeReconciler.Planner *planner.NodeResolver`.

No CRD schema changes.

## Tests

**Planner unit** (`internal/planner/node_update_test.go`):
- `TestBuildRunningPlan_SidecarReady_NoPlan`
- `TestBuildRunningPlan_SidecarNotReady_ReturnsMarkReadyPlan`
- `TestBuildRunningPlan_SidecarUnknown_NoPlan`
- `TestBuildRunningPlan_ImageDriftWinsOverSidecar`
- `TestBuildMarkReadyPlan_FreshIDEveryCall`
- `TestSidecarNeedsReapproval` (all four edge cases)

**Probe unit** (`internal/planner/sidecar_probe_test.go`):
- `TestProbeSidecarHealth_200_SetsReady`
- `TestProbeSidecarHealth_503_SetsNotReady`
- `TestProbeSidecarHealth_NetworkError_SetsUnknown`
- `TestResolvePlan_NilClient_DoesNotProbe`
- `TestResolvePlan_Initializing_DoesNotProbe`

**Reconciler integration** (`internal/controller/node/`):
- `TestReconcile_Running_ProbesAndSetsCondition`
- `TestReconcile_Initializing_SkipsProbe`
- `TestIntegration_PodReplacement_ReissuesMarkReady` — full round-trip: steady-state → flip sidecar to 503 → controller spawns `MarkReady` plan → flip back to 200 → plan reaches Complete and is cleared → `SidecarReady` flips back to True. Asserts mock received a `mark-ready` submission.

All existing tests pass unchanged.

## Verification

- [x] All unit + integration tests pass
- [x] `go vet ./internal/... ./api/...` clean
- [x] lint clean (ginkgolinter, gofmt, modernize, unparam)
- [x] Post-merge: force-delete a release-6-5 validator pod in dev, observe within ~30-60s that the controller spawns a `MarkReady` plan and the pod rejoins consensus without manual `port-forward` intervention.

## Follow-ups

- Platform-side `PrometheusRule` keyed on `seinode_sidecar_ready == 0 for 5m`. Design doc flagged this; implementation belongs in the platform repo's substrate overlay.
- Update `clusters/dev/release-6-5/runbook.md` test 04 to remove the "manually POST mark-ready" step once this ships.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -170,11 +170,11 @@ func main() {
 	//nolint:staticcheck // TODO: migrate to GetEventRecorder (new events API)
 	nodeRecorder := mgr.GetEventRecorderFor("seinode-controller")
 	if err := (&nodecontroller.SeiNodeReconciler{
-		Client:             kc,
-		Scheme:             mgr.GetScheme(),
-		Recorder:           nodeRecorder,
-		Platform:           platformCfg,
-		BuildSidecarClient: buildSidecarClient,
+		Client:   kc,
+		Scheme:   mgr.GetScheme(),
+		Recorder: nodeRecorder,
+		Platform: platformCfg,
+		Planner:  &planner.NodeResolver{BuildSidecarClient: buildSidecarClient},
 		PlanExecutor: &planner.Executor[*seiv1alpha1.SeiNode]{
 			ConfigFor: func(_ context.Context, node *seiv1alpha1.SeiNode) task.ExecutionConfig {
 				return task.ExecutionConfig{
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -170,10 +170,11 @@ func main() {
 	//nolint:staticcheck // TODO: migrate to GetEventRecorder (new events API)
 	nodeRecorder := mgr.GetEventRecorderFor("seinode-controller")
 	if err := (&nodecontroller.SeiNodeReconciler{
-		Client:   kc,
-		Scheme:   mgr.GetScheme(),
-		Recorder: nodeRecorder,
-		Platform: platformCfg,
+		Client:             kc,
+		Scheme:             mgr.GetScheme(),
+		Recorder:           nodeRecorder,
+		Platform:           platformCfg,
+		BuildSidecarClient: buildSidecarClient,
 		PlanExecutor: &planner.Executor[*seiv1alpha1.SeiNode]{
 			ConfigFor: func(_ context.Context, node *seiv1alpha1.SeiNode) task.ExecutionConfig {
 				return task.ExecutionConfig{
--- a/internal/controller/node/controller.go
+++ b/internal/controller/node/controller.go
@@ -108,23 +108,30 @@ func (r *SeiNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		statusDirty = true
 	}
 
-	// Pre-plan: probe the sidecar's Healthz when Running. A 503 here
-	// indicates the mark-ready gate got re-closed by a pod replacement;
-	// the planner picks up the SidecarReady=False condition and spawns a
-	// one-task MarkReady plan. Probe is skipped during Initializing — the
-	// init plan owns sidecar interaction there.
-	if node.Status.Phase == seiv1alpha1.PhaseRunning {
-		if dirty := r.probeSidecarHealth(ctx, node); dirty {
-			statusDirty = true
+	// Resolve or resume plan. ResolvePlan probes the sidecar when the node
+	// is Running (stamping SidecarReady on the node), then either resumes
+	// an active plan or builds a new one based on the node's phase. A
+	// sidecar 503 results in a one-task MarkReady plan.
+	prevSidecar := apimeta.FindStatusCondition(node.Status.Conditions, seiv1alpha1.ConditionSidecarReady)
+	var client task.SidecarClient
+	if r.BuildSidecarClient != nil && node.Status.Phase == seiv1alpha1.PhaseRunning {
+		c, err := r.BuildSidecarClient(node)
+		if err == nil {
+			client = c
 		}
 	}
 
-	// Resolve or resume plan. ResolvePlan either resumes an active plan or
-	// builds a new one based on the node's phase, stamping it onto
-	// node.Status.Plan (and transitioning Pending → Initializing).
 	planAlreadyActive := node.Status.Plan != nil && node.Status.Plan.Phase == seiv1alpha1.TaskPlanActive
-	if err := planner.ResolvePlan(ctx, node); err != nil {
+	if err := planner.ResolvePlan(ctx, node, client); err != nil {
 		return ctrl.Result{}, fmt.Errorf("resolving plan: %w", err)
+	}
+
+	// Emit events on SidecarReady transitions so `kubectl describe` surfaces
+	// them during incidents. The condition itself is the durable signal.
+	r.emitSidecarReadinessEvent(node, prevSidecar)
+	if cur := apimeta.FindStatusCondition(node.Status.Conditions, seiv1alpha1.ConditionSidecarReady); cur != nil &&
+		(prevSidecar == nil || prevSidecar.Status != cur.Status || prevSidecar.Reason != cur.Reason) {
+		statusDirty = true
 	}
 
 	var result ctrl.Result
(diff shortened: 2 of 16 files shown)


internal/controller/node/controller.go
```diff
@@ -99,6 +108,17 @@ func (r *SeiNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	// Pre-plan: probe the sidecar's Healthz when Running. A 503 here
+	// indicates the mark-ready gate got re-closed by a pod replacement;
+	// the planner picks up the SidecarReady=False condition and spawns a
+	// one-task MarkReady plan. Probe is skipped during Initializing — the
+	// init plan owns sidecar interaction there.
+	if node.Status.Phase == seiv1alpha1.PhaseRunning {
+		if dirty := r.probeSidecarHealth(ctx, node); dirty {
+			statusDirty = true
+		}
+	}
```
bdchatham: I was thinking we'd put this in the planner so that when you call planner.ForNode that this would be checked.

internal/controller/node/controller.go
```diff
@@ -116,6 +136,10 @@ func (r *SeiNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		// (and any conditions) before side effects occur.
 		statusDirty = true
 		result = planner.ResultRequeueImmediate
+		if isMarkReadyReapplyPlan(node.Status.Plan) {
```
bdchatham: Why do we need this?

internal/controller/node/controller.go
bdchatham: Moved in 41e3109. `probeSidecarHealth` now lives in `internal/planner/sidecar_probe.go` and is called from `ResolvePlan` (which takes a `task.SidecarClient`). The reconciler just wires the client through; drift-detection + mitigation are co-located in the planner.

internal/controller/node/controller.go
bdchatham: Fair call — `SidecarReadinessLost` already signals the state transition and the plan is visible in `status.plan`. Dropped in 41e3109.

internal/planner/planner.go
bdchatham, line 79: It seems like we should just create the sidecar client inside the planner. The planner then becomes a struct that has this function to generate the client. This function serves as the means to inject behavior into the planner for prod deployment and testing.

cmd/main.go
```diff
@@ -170,10 +170,11 @@ func main() {
 	//nolint:staticcheck // TODO: migrate to GetEventRecorder (new events API)
 	nodeRecorder := mgr.GetEventRecorderFor("seinode-controller")
 	if err := (&nodecontroller.SeiNodeReconciler{
```
bdchatham, line 172: We will pass in the planner construct here and that will have the sidecar client function inside it.

internal/controller/node/controller.go
```diff
@@ -99,14 +108,32 @@ func (r *SeiNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	// an active plan or builds a new one based on the node's phase. A
+	// sidecar 503 results in a one-task MarkReady plan.
+	prevSidecar := apimeta.FindStatusCondition(node.Status.Conditions, seiv1alpha1.ConditionSidecarReady)
+	var client task.SidecarClient
+	if r.BuildSidecarClient != nil && node.Status.Phase == seiv1alpha1.PhaseRunning {
+		c, err := r.BuildSidecarClient(node)
+		if err == nil {
+			client = c
+		}
+	}
```
bdchatham: If this is in the planner struct, I think we can push this into that code path and keep the reconciliation cleaner.

internal/planner/metrics.go
```diff
@@ -36,6 +41,60 @@ func init() {
...
+	return nil
+}
+
+func splitKey(key string) (ns, name string) {
+	for i := 0; i < len(key); i++ {
+		if key[i] == '/' {
+			return key[:i], key[i+1:]
+		}
+	}
+	return "", key
```
bdchatham: Instead of doing this split in the key, can we add a label for namespace? The metric should be just the metric name. Labels are idiomatic for adding contextual details.

bdchatham: Good call. Switched to a `nodeKey{namespace, name}` struct key — no more concat/split. Namespace and name stay as metric attributes on emit, which was already correct. Fixed in 395d25b.

PR #135, opened 2026-04-27
docs: LLD for SeiNode validator signing-key feature

bdchatham:
## Summary

Low-level design for adding `validator.signingKey` to the `SeiNode` CRD, enabling migration of an existing external validator identity onto a Kubernetes-managed validator without double-signing risk.

Companion to [`.tide/validator-migration.md`](https://github.com/sei-protocol/sei-k8s-controller/blob/main/.tide/validator-migration.md) (direction doc) and follows the shape discipline established by [`docs/design-seinode-import-volume-lld.md`](https://github.com/sei-protocol/sei-k8s-controller/blob/main/docs/design-seinode-import-volume-lld.md).

## What it locks down

- **API shape:** `SigningKey *SigningKeySource` with one v1 variant `Secret *SecretSigningKeySource`. Discriminated union with exactly-one CEL rule shipped in v1 so future variants (TMKMS, Horcrux, remote signer, Vault) are additive — no breaking change required.
- **Mount mechanics:** one Secret-backed volume; `subPath` mount of `priv_validator_key.json` into the seid container at `$SEI_HOME/config/`. `subPath` is the safety property — kubelet does not auto-refresh `subPath` mounts, so a `kubectl edit secret` cannot hot-swap the consensus key under a running validator. Rotating the key requires a deliberate pod restart paired with on-chain `MsgEditValidator`.
- **Bootstrap-Job invariant:** the bootstrap-Job pod-spec generator is deliberately untouched — bootstrap pods are physically incapable of signing because they have no key material. A load-bearing comment pins this so a future refactor cannot accidentally couple the two pod-spec generators.
- **Pre-flight validation:** new `validate-signing-key` task surfaces missing/malformed Secrets via a `SigningKeyReady` Status condition mirroring the `ImportPVCReady` reason taxonomy.
- **Cutover relies on existing NodeUpdate plan:** patching `signingKey` onto a pre-synced SeiNode triggers the standard NodeUpdate StatefulSet-diff path; no new "validator-promotion" plan needed.

## Notable scope decision

`priv_validator_state.json` is **not** injected by the controller. CometBFT auto-creates the file with zero state on first start; the cutover runbook's hard-halt-old-validator + wait-≥M-blocks-for-chain-advance is the operational protection against cutover-boundary re-orgs. The state file is operational data (height/round/step + most-recent signature, all of which are public on-chain), not secret material — putting it through controller-mediated injection would be over-classification. Documented in §8 and §11.

## Explicitly deferred (§11)

TMKMS / Horcrux / remote signer variants, automated cutover orchestration, double-sign detection, sentry-node topology, consensus-key rotation, cross-namespace Secret references, HSM integration, unsetting `SigningKey` after it's been set.

## Test plan

This PR contains design only — no code changes. Test plan applies when implementation lands:

- [x] CRD validation: `signingKey` accepted with `secret` variant, rejected without; `secretName` immutable once set
- [x] Pod spec: production StatefulSet has `signing-key` Secret volume with `subPath` mount on seid container; sidecar has no signing mount
- [x] Bootstrap-Job invariant: `task.GenerateBootstrapJob` produces a Job with no signing volume even when `SigningKey` is set on the SeiNode
- [x] `validate-signing-key` task: transient on missing Secret, terminal on malformed key, completes on valid Secret
- [x] Status condition: `SigningKeyReady` set with reasons matching §5 matrix; `ObservedGeneration` updated
- [x] Cutover flow: SeiNode applied without `SigningKey` reaches Running as observer; patching `SigningKey` in triggers pod restart with key mounted
- [x] Finalizer: deleting SeiNode does not delete the referenced Secret

### DIFF
--- a/docs/design-seinode-validator-signing-key-lld.md
+++ b/docs/design-seinode-validator-signing-key-lld.md
@@ -27,14 +27,18 @@ A single new optional sub-struct is added to `ValidatorSpec` in `api/v1alpha1/va
 // api/v1alpha1/validator_types.go
 
 type ValidatorSpec struct {
+    // Snapshot, GenesisCeremony — existing fields, both optional.
+    // +optional
     Snapshot        *SnapshotSource             `json:"snapshot,omitempty"`
+    // +optional
     GenesisCeremony *GenesisCeremonyNodeConfig  `json:"genesisCeremony,omitempty"`
 
     // SigningKey declares the source of this validator's consensus signing
     // key (priv_validator_key.json). When omitted, the node runs as a
     // non-signing observer — suitable for pre-sync (Phase 1 of the
     // validator-migration runbook) or for genesis-ceremony bootstraps that
-    // produce keys on-cluster.
+    // produce keys on-cluster. Mutually exclusive with GenesisCeremony
+    // (cross-field check in the validator planner; see §6).
     // +optional
     SigningKey *SigningKeySource `json:"signingKey,omitempty"`
 }
--- a/docs/design-seinode-validator-signing-key-lld.md
+++ b/docs/design-seinode-validator-signing-key-lld.md
@@ -1,0 +1,430 @@
+# Design: SeiNode — Validator Signing Key (LLD)
+
+**Status:** Draft / LLD
+**Date:** 2026-04-27
+**Tracks:** validator migration from sei-infra (EC2) → sei-k8s-controller
+**Related:** [`.tide/validator-migration.md`](../.tide/validator-migration.md), [`docs/design-seinode-import-volume-lld.md`](design-seinode-import-volume-lld.md), [`docs/design/composable-genesis.md`](design/composable-genesis.md)
+
+This LLD specifies how an existing external validator identity is mounted onto a `SeiNode` so the node, once synced to the chain tip, takes over signing from a previously-running validator instance with zero risk of double-signing.
+
+The companion direction doc (`.tide/validator-migration.md`) sets the threat model and migration phases. This LLD fixes the API shape, mount mechanics, validation lifecycle, and v1 scope. No new keystore variants, no remote-signer integration, no automated cutover orchestration — those are explicitly deferred (§11).
+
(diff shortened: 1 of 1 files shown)


docs/design-seinode-validator-signing-key-lld.md
```diff
@@ -0,0 +1,430 @@
...
+## 1. CRD schema changes
+
+A single new optional sub-struct is added to `ValidatorSpec` in `api/v1alpha1/validator_types.go`. The shape is a discriminated union with one v1 variant; future variants (TMKMS, remote signer, Vault) slot in as sibling pointer fields.
+
+```go
+// api/v1alpha1/validator_types.go
+
+type ValidatorSpec struct {
+    Snapshot        *SnapshotSource             `json:"snapshot,omitempty"`
+    GenesisCeremony *GenesisCeremonyNodeConfig  `json:"genesisCeremony,omitempty"`
```
bdchatham, line 34: Does this become optional?

bdchatham: Yes — `GenesisCeremony` is already optional in `api/v1alpha1/validator_types.go` today (set by the SeiNodeDeployment controller for genesis ceremonies, unset otherwise). My snippet elided the existing `+optional` markers for brevity, which made it ambiguous. Pushed 160b827 to make the snippet faithful: `Snapshot`, `GenesisCeremony`, and `SigningKey` are all `+optional`. The cross-field rule (§6) enforces `SigningKey` ↔ `GenesisCeremony` as mutually exclusive — so a validator is in one of three states:

1. `GenesisCeremony` set, `SigningKey` unset — fresh-chain bootstrap, generates keys on-cluster
2. `SigningKey` set, `GenesisCeremony` unset — joining an existing chain with imported keys (the migration use case)
3. Both unset — joining as a non-signing observer (Phase 1 pre-sync, before SigningKey is patched in)

`Snapshot` is independently optional in all three cases.

PR #136, opened 2026-04-28
feat: validator signing-key configuration on SeiNode

bdchatham:
## Summary

**Migrate Sei validators from EC2 to Kubernetes without double-signing.** This PR adds `spec.validator.signingKey.secret.secretName` to the `SeiNode` CRD so an operator can deploy a K8s validator that takes over signing from a stopped EC2 instance using its existing consensus identity.

Implements [`docs/design-seinode-validator-signing-key-lld.md`](https://github.com/sei-protocol/sei-k8s-controller/blob/main/docs/design-seinode-validator-signing-key-lld.md) (PR #135). Direction doc: [`.tide/validator-migration.md`](https://github.com/sei-protocol/sei-k8s-controller/blob/feat/validator-signing-key/.tide/validator-migration.md).

## Scope

v1 supports **single-shot deployment**: deploy a `SeiNode` with `validator.signingKey.secret.secretName` set from creation. The bootstrap Job runs without the Secret (load-bearing safety property — bootstrap pods are physically incapable of signing); the production StatefulSet starts with the Secret mounted, seid catches up to chain tip and starts signing.

**Mid-life patching is NOT supported** — adding `signingKey` to a Running validator is a no-op today because `buildRunningPlan` only detects image drift. Documented in LLD §11 with the implementation sketch for the deferred zero-downtime variant.

## What's in this PR

**CRD additions** (`api/v1alpha1/`)
- `SigningKeySource` discriminated union with one v1 variant `Secret *SecretSigningKeySource`. Future TMKMS / RemoteSigner / Vault variants slot in additively under the same exactly-one CEL rule.
- `secretName` immutable (`self == oldSelf`).
- `ConditionSigningKeyReady` + `Validated` / `NotReady` / `Invalid` reason taxonomy (mirrors `ImportPVCReady`).

**Pod-spec mutation** (`internal/noderesource/`, production StatefulSet only)
- One Secret-backed volume scoped to `priv_validator_key.json` (defaultMode 0400, items-filtered).
- `subPath` mount on the seid main container at `/sei/config/priv_validator_key.json`. `subPath` is the safety property — kubelet does not auto-refresh `subPath` mounts, so a `kubectl edit secret` cannot hot-swap the consensus key under a running seid.
- Sidecar container has no signing mount.

**Bootstrap-Job safety invariant** (`internal/task/bootstrap_resources.go`)
- `assertNoSigningKeyOnBootstrapPod` runtime guard fails Job generation if the bootstrap pod-spec ever references the validator's signing-key Secret. Belt-and-suspenders against a future refactor that couples the volume helpers; the existing test `TestTaskGenerateBootstrapJob_NeverHasSigningKeyVolume` is the static-analysis layer.

**Pre-flight validation** (`internal/task/validate_signing_key.go`)
- New `validate-signing-key` task: existence, deletionTimestamp, key data present, JSON parse, Tendermint shape (address, pub_key.{type,value}, priv_key.{type,value}). Direct condition mutation from the task (matches the merged `ensure-data-pvc` precedent).
- Inserted into both `buildBootstrapPlan` (after `EnsureDataPVC`, before `DeployBootstrapSvc`) and `buildBasePlan` (after `EnsureDataPVC`, before `ApplyStatefulSet`) when SigningKey is set.

**Validator planner cross-field validation** — `SigningKey` is mutually exclusive with `GenesisCeremony`.

**RBAC** — `+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch` on the SeiNode reconciler.

## Notable scope decisions

- `priv_validator_state.json` is **not** injected. CometBFT auto-creates it on first start. The runbook's "wait M blocks past last-signed height before deploying" provides the slashing-protection envelope.
- Mid-life SigningKey patch on a Running validator is **deferred**. v1 ships single-shot deployment only.

## What's deferred (LLD §11)

Mid-life SigningKey patch (drift detection); TMKMS / Horcrux / RemoteSigner variants; automated cutover orchestration; double-sign detection; sentry-node topology; consensus-key rotation; cross-namespace Secret references; HSM integration.

## Test plan

- [x] `make test` — all packages green (`internal/task` 41.9% / `internal/noderesource` 96.2% / `internal/planner` 69.3% / `internal/controller/node` 85.6%)
- [x] `make lint` — 0 issues
- [x] `make manifests generate` — regenerated CRD YAML, deepcopy, role.yaml
- [x] Task unit tests (9): valid Secret; missing Secret transient; terminating Secret; missing data key; empty data key; malformed JSON; missing address; missing pub_key; **`tendermintValidatorKey` serdes round-trip fixture**
- [x] Pod-spec tests (4): volume present, `subPath` mount on seid, sidecar mount absent, regression guard for unset SigningKey
- [x] Bootstrap-Job invariant test: pod-spec must NEVER carry signing volume even when SigningKey is set on the SeiNode (LLD §3); plus the new runtime guard `assertNoSigningKeyOnBootstrapPod`
- [x] Plan-builder tests (5): Validate cross-field rules; plan ordering for bootstrap path; plan ordering for base path; absence when SigningKey unset
- [x] Controller integration tests (4): SigningKeyReady=True on valid Secret; missing-Secret transient → applied → converges; malformed-Secret → plan Failed + condition False/Invalid; SeiNode deletion preserves the referenced Secret

## Operational dry-run before first cutover

The runbook in `.tide/validator-migration.md` is the long pole. Plan: walk the single-shot deployment against a throwaway arctic-1 testnet identity (lowest-staked validator first), confirm the M-block wait works, confirm seid auto-creates `priv_validator_state.json` and starts signing past the wait window. Document downtime envelope before applying to pacific-1.


internal/planner/bootstrap.go
```diff
@@ -50,6 +50,16 @@ func buildBootstrapPlan(
 		return nil, err
 	}
 
+	// Phase 0.5: fail fast on a misconfigured Secret before spinning up the
+	// bootstrap Job. The Secret never reaches the bootstrap pod (safety
+	// invariant on task.GenerateBootstrapJob).
```
bdchatham: Can you delete this comment?

internal/task/bootstrap_resources.go
```diff
@@ -38,6 +38,11 @@ func BootstrapLabels(node *seiv1alpha1.SeiNode) map[string]string {
 
 // GenerateBootstrapJob creates the batch Job that runs seid with --halt-height
 // to populate a PVC before the StatefulSet takes over.
+//
+// SAFETY INVARIANT: the bootstrap pod-spec must never carry ValidatorSpec.SigningKey
+// volumes. Bootstrap pods are physically incapable of signing because no key
+// material is on their filesystem; do not extract a shared volume helper
+// with noderesource.buildNodePodSpec that would couple the two paths.
```
bdchatham: Is there any programmatic validation we can add instead of this comment?

internal/task/validate_signing_key.go
```diff
@@ -0,0 +1,147 @@
...
+	Address string `json:"address"`
+	PubKey  struct {
+		Type  string `json:"type"`
+		Value string `json:"value"`
+	} `json:"pub_key"`
+	PrivKey struct {
+		Type  string `json:"type"`
+		Value string `json:"value"`
+	} `json:"priv_key"`
+}
```
bdchatham: Any tests we can add to ensure this model fits? Maybe add a fixture to ensure it serdes properly. Also, can we put this at the top of the file?

internal/planner/bootstrap.go
bdchatham: Done in eea6b1a — comment removed, the task ordering speaks for itself.

internal/task/bootstrap_resources.go
bdchatham: Done in eea6b1a. Replaced the prose with a runtime guard `assertNoSigningKeyOnBootstrapPod` (`internal/task/bootstrap_resources.go`): after `buildBootstrapPodSpec` returns, the function scans the resulting pod-spec for any volume that references `validator.signingKey.secret.secretName` and returns an error if found. A future refactor that accidentally couples the volume helpers will fail closed at Job-generation time. The existing test `TestTaskGenerateBootstrapJob_NeverHasSigningKeyVolume` remains as the static-analysis layer.

internal/task/validate_signing_key.go
bdchatham: Done in eea6b1a:
- Moved `tendermintValidatorKey` to the top of the file, just under the constants.
- Added `TestTendermintValidatorKey_FixtureRoundTrip` that unmarshals the `validKeyJSON` fixture into the type, asserts each required field is populated, then re-marshals and confirms the on-wire shape contains all five required fields (address, pub_key.type/value, priv_key.type/value). Locks the model against a future cosmos-sdk rename or field removal.

Issue #137, opened 2026-04-28
Detect spec drift on Running nodes for mid-life SigningKey patch (and future validator mode switch)


bdchatham:
## Background

PR #136 shipped `validator.signingKey.secret.secretName` for the single-shot validator deployment use case — deploy a `SeiNode` with SigningKey set from creation and the controller mounts the Secret on the production StatefulSet pod.

What it explicitly does **not** support: patching `SigningKey` onto an already-Running validator. `buildRunningPlan` (`internal/planner/planner.go:621-628`) only detects image drift today, so a `kubectl patch seinode --patch '{"spec":{"validator":{"signingKey":...}}}'` against a Running node is a silent no-op — the pod doesn't restart, the Secret never mounts, the validator never starts signing. Documented in [LLD §11](https://github.com/sei-protocol/sei-k8s-controller/blob/main/docs/design-seinode-validator-signing-key-lld.md#11-what-this-lld-does-not-cover).

### Primary use case — zero-downtime migration cutover

The single-shot deployment in PR #136 trades cutover downtime (bootstrap-Job sync + StatefulSet catch-up) for implementation simplicity. For arctic-1 testnet this is fine; for pacific-1 the downtime envelope may not be acceptable. The original LLD §8 design was a two-phase cutover:

1. Deploy `SeiNode` without `signingKey` → bootstrap + sync as a non-signing observer (zero risk, no downtime on the old EC2 validator)
2. Stop EC2 → wait M blocks → patch `signingKey` in → pod re-rolls with Secret mounted → seid signs

Phase 2 requires drift detection. Without it, the operator's choices are (a) accept the single-shot downtime, or (b) delete-and-recreate the SeiNode (which deletes the data PVC unless `dataVolume.import` is wired, costing the bootstrap+sync work).

### Future use case — validator mode switch

A separate, larger ask: convert a Running `fullNode` SeiNode into a `validator` SeiNode by patching the spec (e.g., `spec.fullNode → spec.validator` with a `signingKey`). Same drift-detection mechanic, larger pod-spec surface (mode string in seid config, port configuration, probes). Out of scope for v1 of this issue, but the underlying machinery is identical — listing here so the design accommodates it additively.

## Proposed mechanic

Reuse the existing NodeUpdate plan shape (`buildNodeUpdatePlan` at `internal/planner/planner.go:656`):

[...]

PR #140, opened 2026-04-28
feat: validator nodeKey configuration on SeiNode

### LINKED ISSUE #138: Add nodeKey field to ValidatorSpec for stable P2P node identity
## Problem

PR #136 lets a validator import its consensus key (`priv_validator_key.json`) via `validator.signingKey`, but does not let it import its P2P node key (`node_key.json`). When the K8s validator boots, seid auto-generates a fresh `node_key.json` on the data PVC, giving the validator a new node ID. For the arctic-1 / pacific-1 EC2 → K8s migration this means every migrated validator presents itself to the network at a new P2P address — a "known validator" with a "new node ID" — which is exactly the impersonation signal topology trackers and peer-reputation systems flag. It also silently breaks any sentry-node topology that pinned the validator's node ID in `private_peer_ids`, and breaks `persistent_peers` entries in any operator's config that references the validator by old ID.

## Impact

- **Migration breakage** — every migrated validator looks like a new entity at the P2P layer; `persistent_peers` entries across the network become stale.
- **Sentry topologies break silently** — sentries pinning the validator's old node ID in `private_peer_ids` lose the gossip shielding that protects the validator.
- **Peer-reputation drift** — libp2p connection persistence and reputation are keyed off node ID, so a regen costs accumulated peer goodwill.
- **Operational discontinuity** — any topology tracker / network-observer dashboard that maps node IDs to validator monikers shows a discontinuity at cutover, complicating incident debugging precisely when downtime is most costly.

## Relevant experts

- `kubernetes-specialist` — pod-spec mutation parallel to the existing signingKey path; dual-pod (bootstrap + production) mount semantics
- `platform-engineer` — API shape and the `signingKey` ↔ `nodeKey` field-coupling rule
- `security-specialist` — whether `node_key.json` deserves Secret-grade handling or if a ConfigMap variant is acceptable (non-slashing but identity-bearing)

## Proposed approach

API addition (`api/v1alpha1/validator_types.go`):

```go
type ValidatorSpec struct {
    Snapshot        *SnapshotSource           `json:"snapshot,omitempty"`
    GenesisCeremony *GenesisCeremonyNodeConfig `json:"genesisCeremony,omitempty"`
    SigningKey      *SigningKeySource          `json:"signingKey,omitempty"`
    NodeKey         *NodeKeySource             `json:"nodeKey,omitempty"`  // new
}

type NodeKeySource struct {
    Secret *SecretNodeKeySource `json:"secret,omitempty"`
}

type SecretNodeKeySource struct {
    SecretName string `json:"secretName"`  // Secret data key: "node_key.json"
}
```

Cross-field coupling: `XValidation: has(self.signingKey) == has(self.nodeKey)` on `ValidatorSpec`. Validators get both or neither in v1; relax independently if non-signing observer use cases want stable node IDs later.

**Pod-spec mutation differs from `signingKey` in one important way:** `nodeKey` mounts on **both the bootstrap Job pod AND the production StatefulSet pod**, unlike `signingKey` (production-only). Rationale: `node_key.json` is P2P identity, not consensus identity — not slashing-relevant. Mounting on bootstrap means the validator presents a consistent node ID to peers throughout bootstrap + sync. The "bootstrap pod cannot sign" safety invariant is unaffected because `signingKey` remains production-only.

Mount: `$SEI_HOME/config/node_key.json`, `subPath`, `defaultMode: 0400`, `ReadOnly: true`. Same shape as `signingKey` except the volume is added to both `buildNodePodSpec` and `buildBootstrapPodSpec`.

New pre-flight task `validate-node-key` mirroring `validate-signing-key`: existence, deletionTimestamp, data key present, JSON parse, Tendermint node-key shape (`priv_key.{type,value}`). Same `Validated`/`NotReady`/`Invalid` reason taxonomy. Surfaces via a new `NodeKeyReady` Status condition.

## Acceptance criteria

- [ ] `nodeKey` field added to `ValidatorSpec` with `Secret` variant; XValidation rule couples it with `signingKey`
- [ ] Pod-spec generators mount the Secret on both bootstrap and production pods (subPath, mode 0400)
- [ ] `validate-node-key` task implemented and inserted into validator plans alongside `validate-signing-key`
- [ ] `NodeKeyReady` Status condition with the standard reason taxonomy
- [ ] Integration test: validator with `nodeKey` set boots both bootstrap pod AND production pod with `node_key.json` mounted at the expected path
- [ ] Bootstrap-Job invariant test extended: still no signing-key volume, but DOES include node-key volume when `nodeKey` is set
- [ ] `assertNoSigningKeyOnBootstrapPod` stays unchanged (it scans for the signingKey Secret name only)
- [ ] LLD updated (or new mini-LLD added) documenting the dual-pod mount and the safety-difference rationale

## Out of scope

- **Rotating `node_key.json`.** Like `priv_validator_key.json`, `secretName` is immutable; rotation requires delete-and-recreate.
- **`nodeKey` support on `fullNode` / `archive` / `replayer` modes.** Defer until a concrete operator need surfaces.
- **Mid-life patching of `nodeKey` onto a Running validator.** Requires drift detection — see #137.
- **ConfigMap variant for `nodeKey`.** Defer; `secretName` is consistent with `signingKey` for v1, and a ConfigMap variant can be added additively later.

## References

- PR #136 — single-shot deployment for `SigningKey`
- Issue #137 — drift detection for mid-life patching (would extend to `nodeKey`)
- [LLD: docs/design-seinode-validator-signing-key-lld.md](https://github.com/sei-protocol/sei-k8s-controller/blob/main/docs/design-seinode-validator-signing-key-lld.md)

bdchatham:
## Summary

Closes #138.

Adds `spec.validator.nodeKey` to `SeiNode` so a migrating validator can bring its existing P2P node key (`node_key.json`) onto K8s instead of regenerating one. Stable node IDs matter for `persistent_peers` entries elsewhere in the network, sentry-node `private_peer_ids` configurations, and libp2p peer reputation — a "known validator coming up with a new node ID" is exactly the impersonation signal topology trackers flag.

`nodeKey` is **required when `signingKey` is set** — validators get both or neither in v1.

## Notable design decision: production-only mount

Per coral-team review and Brandon's call: `nodeKey` mounts on the **production StatefulSet pod only**, not the bootstrap Job pod. The bootstrap Job uses the ephemeral node ID generated by `seid init` for its short-lived sync window; the validator's permanent node ID first appears on the network when the production pod starts.

This avoids attributing the bootstrap pod's lifecycle drama (snapshot-restore failures, halt-on-130, OOMs) to the validator's permanent libp2p reputation. The "validator's first appearance = production pod" framing is also cleaner from a peer-graph perspective for fresh deployments.

If a future use case needs continuity-through-bootstrap, that's an additive opt-in field — no API churn required.

## What's in this PR

**CRD additions** (`api/v1alpha1/`)
- `NodeKeySource` discriminated union with one v1 variant `Secret *SecretNodeKeySource`. Same exactly-one CEL pattern as `SigningKeySource`.
- `secretName` immutable (`self == oldSelf`).
- Two `XValidation` rules on `ValidatorSpec`:
  - `has(self.signingKey) == has(self.nodeKey)` — couples the two fields
  - `self.signingKey.secret.secretName != self.nodeKey.secret.secretName` — defense-in-depth: rejects packing both keys in one Secret (would collapse the bootstrap-pod trust boundary)
- `ConditionNodeKeyReady` + `ReasonNodeKey{Validated,NotReady,Invalid}` taxonomy.

**Pod-spec mutation** (`internal/noderesource/`)
- `nodeKeyVolumes` / `nodeKeyMounts` / `nodeKeySecretSource` helpers parallel to the signingKey trio. Pure copy per coral consensus — the safety stories are different (signingKey is production-only via load-bearing comment; nodeKey is benign-on-either-pod), and forcing one helper to carry both invariants makes the code less greppable.
- `subPath` mount at `/sei/config/node_key.json` on seid main container; sidecar untouched.

**Bootstrap-Job safety** (`internal/task/bootstrap_resources.go`)
- Bootstrap pod-spec **deliberately untouched** — production-only mount.
- Existing `assertNoSigningKeyOnBootstrapPod` runtime guard works as-is for nodeKey because the cross-field rule rejects shared-secretName configurations, so the bootstrap pod can't accidentally end up with the signing-key Secret via a packed-Secret.
- Regression-guard test `TestTaskGenerateBootstrapJob_NeverHasIdentityVolumes` renamed and extended to cover BOTH `signing-key` and `node-key` volume names + Secret names.

**Pre-flight validation** (`internal/task/validate_node_key.go`)
- New task mirrors `validate-signing-key`. Tendermint shape check is `priv_key.{type,value}` only (node_key.json has no `address` or `pub_key`).
- Direct condition mutation from the task, same `ensure-data-pvc`-style precedent.
- Inserted into both `buildBootstrapPlan` and `buildBasePlan` after `validate-signing-key` when `nodeKey` is set.

**Validator planner cross-field** (`internal/planner/validator.go`)
- Go-side mirror of CEL: nodeKey required when signingKey is set; secretNames must be distinct.

## Test plan

- [x] `make test` — all packages green
- [x] `make lint` — 0 issues
- [x] `make manifests generate` — regenerated CRD YAML, deepcopy
- [x] Task unit tests (8): valid Secret, missing Secret transient, missing data key terminal, malformed JSON terminal, missing priv_key terminal, missing-then-applied convergence, **`tendermintNodeKey` serdes round-trip fixture**
- [x] Pod-spec tests (5): volume present, subPath mount on seid, sidecar mount absent, regression guard for unset, both-mounts-coexist with distinct paths
- [x] Bootstrap-Job invariant test: pod must NEVER carry signing-key OR node-key volumes/Secrets even when both are set on the SeiNode (unified safety guard)
- [x] Plan-builder tests: extended ordering assertions (`pvc < signing < nodeKey < statefulset`) for both bootstrap and base paths; new Validate cases for signingKey-without-nodeKey, nodeKey-without-signingKey, shared-secretName, and the existing happy paths now pair both keys

## What's deferred

- Mid-life patching of nodeKey onto a Running validator (covered by drift-detection issue #137).
- nodeKey support on fullNode / archive / replayer modes.
- ConfigMap variant for nodeKey (operators don't need it; node_key.json is treated as Secret-grade for handling consistency with signingKey).
- Continuity-through-bootstrap (mounting nodeKey on bootstrap pod) — additive opt-in field if/when concrete operator need surfaces.


internal/noderesource/noderesource.go
```diff
@@ -512,3 +516,53 @@ func signingKeySecretSource(node *seiv1alpha1.SeiNode) *seiv1alpha1.SecretSignin
...
+
+const (
+	nodeKeyVolumeName = "node-key"
+	nodeKeyDataKey    = "node_key.json"
+)
+
+// nodeKeyVolumes mounts only on the production StatefulSet pod. The bootstrap
+// Job uses an ephemeral node ID generated by `seid init` so its lifecycle
+// drama (snapshot-restore failures, halt-on-130) doesn't pollute the
+// validator's permanent libp2p reputation.
```
bdchatham: Can we make this more concise?

internal/noderesource/noderesource.go
```diff
@@ -512,3 +516,53 @@ func signingKeySecretSource(node *seiv1alpha1.SeiNode) *seiv1alpha1.SecretSignin
 	}
 	return node.Spec.Validator.SigningKey.Secret
 }
+
+const (
+	nodeKeyVolumeName = "node-key"
+	nodeKeyDataKey    = "node_key.json"
+)
```
bdchatham: Can you move these to the top of the file?

internal/noderesource/noderesource.go
bdchatham: Done in d6c4a57. Trimmed to two lines — the libp2p-reputation rationale belongs in the LLD, not at the helper:
```go
// nodeKeyVolumes mounts the node-key Secret on the production pod only.
// The bootstrap Job uses an ephemeral node ID generated by `seid init`.
```

internal/noderesource/noderesource.go
bdchatham: Done in d6c4a57. Hoisted both pairs (`signingKey*` and `nodeKey*`) into the top-of-file const block alongside `dataDir` and `defaultSidecarImage` so all file-level constants live in one place.

Issue #147, opened 2026-04-29
Verify all controller status patches use optimistic concurrency, not blind SSA


bdchatham:
## Problem

Plan-creation idempotency under fast double-reconcile depends on the executor's single-patch model + status-conflict retry. If status patches are blind SSA (no `resourceVersion` check), two near-simultaneous reconciles can both observe `status.plan == nil`, both build a plan, and the second's `Patch` silently overwrites the first — losing the first plan's ID and any in-flight state.

The ValidationRun LLD (PR #143) names this as an implementation invariant for the new ValidationOrchestrationReconciler and ValidationLoadGenerationReconciler controllers. Existing controllers (`SeiNodeReconciler`, `SeiNodeDeploymentReconciler`) appear to use optimistic concurrency, but this hasn't been audited as a code-review checklist item — it's a property the codebase relies on without enforcement.

### Primary use case
Any controller in this repo that builds plans via `ResolvePlan`-style logic. Today: SeiNode, SND. Tomorrow: ValidationRun (orchestration + load generation).

### Cost of not addressing
A single contributor mistakenly using `client.Status().Update(ctx, obj)` (or `Patch` without resourceVersion) instead of `client.Status().Patch(ctx, obj, client.MergeFrom(original))` introduces a race that's hard to surface in tests and easy to miss in code review. The race is rare but real, and corrupts plan state when it triggers.

## Relevant experts

- `kubernetes-specialist` — controller-runtime status-patch idioms, optimistic concurrency

## Proposed approach

Three-part fix:

**1. Audit existing controllers.** Verify all status patches in `internal/controller/{node,nodedeployment,...}` use optimistic concurrency. Specifically:
- `client.Status().Patch(ctx, obj, client.MergeFrom(original))` — OK
- `client.Status().Update(ctx, obj)` — verify resourceVersion on `obj` matches a recent fetch
- Blind SSA without resourceVersion — fix

**2. Add a code-review checklist item.** Document in CONTRIBUTING.md (or repo equivalent) that all status patches must be optimistic-concurrency checked.

**3. Add an integration test.** Exercise double-reconcile contention on a single resource — fire two `Reconcile` calls in close succession, verify only one plan creation succeeds and the other observes-and-skips.

## Acceptance criteria

- [ ] Audit completed for SeiNodeReconciler, SeiNodeDeploymentReconciler, and any other planner consumer
[...]

Issue #146, opened 2026-04-29
Decide on TaskPlan.TargetPhase generalization vs per-controller phase ownership


bdchatham:
## Problem

`TaskPlan.TargetPhase` at `api/v1alpha1/seinode_types.go:214` is typed as `SeiNodePhase`. The ValidationRun controllers being designed in PR #143 use a different phase enum (`Pending|Running|Succeeded|Failed|Error|Cancelled`) that is **not** assignable to `SeiNodePhase`.

Per the LLD's resolved decisions, v1 `ValidationOrchestrationReconciler` and `ValidationLoadGenerationReconciler` **ignore `TargetPhase`** and own phase transitions in their own terminal-plan handler (the `finalize` task computes Phase + Verdict from conditions + `status.workloadExitCode` + `status.rules[]`). This is a v1 dodge that lets the planner-package type stay unchanged.

### Why this is a one-way door

If we leave `TaskPlan.TargetPhase` typed as `SeiNodePhase`, future controllers that want to use the planner's phase-transition machinery (mirroring the SeiNode/SND patterns) will hit the same coupling and either reimplement the dodge or force a generalization that affects all existing planner consumers.

When `ValidationSequenceReconciler` ships in v2 — and it has multiple step states that map to natural phase transitions — the dodge gets harder to justify. v2 is where this resolution gets forced.

### Three resolution paths

1. **Defer (LLD's current recommendation).** v1 ValidationRun ignores `TargetPhase`; revisit when a third controller needs the abstraction.
2. **Generalize now.** Change `TaskPlan.TargetPhase` from `SeiNodePhase` to `string` (with per-controller validation). One-time touch across SeiNode/SND/ValidationRun controllers; opens the door for any future controller.
3. **Per-controller plan types.** Each controller defines its own `TaskPlan` type with its own typed `TargetPhase`. Most invasive; doesn't reuse the shared executor code.

## Relevant experts

- `kubernetes-specialist` — planner-package architecture, type design across multiple controllers

## Proposed approach

**Track this as a deferred decision.** Per the ValidationRun LLD's stance:
- v1 ships with the dodge (ValidationRun controllers ignore `TargetPhase`).
- This issue documents the decision-trigger: when the third planner consumer (most likely `ValidationSequenceReconciler`) needs phase transitions through plan completion, re-evaluate paths 2 and 3.
- Path 2 (generalization) is the lean if/when the trigger fires.

[...]

Issue #144, opened 2026-04-29
SeiNode readiness probe should include catching_up=false on full-node-role nodes


bdchatham:
## Problem

`SeiNodeDeployment.status.phase=Ready` (computed at `api/v1alpha1/seinodedeployment_types.go:188`) is satisfied when child node `phase=Running` plus `ConditionNodesReady`, but does **not** include "all full nodes have caught up" (`catching_up=false`). The `AwaitNodesCaughtUp` task exists at `internal/task/deployment.go:46-51` and is invoked during hard-fork rollouts, but not during initial bring-up.

For genesis-bootstrap chains the gap is essentially zero — validators produce blocks from height 0. For full-node fleets that block-sync from validators or external state, there is a real gap between `phase=Ready` (pods running) and "RPC is actually serving caught-up data."

### Primary use case — multi-consumer reliability

This affects three consumers, none of which is currently blocked but all of which would benefit:

- **ValidationRun** (PR #143): the workload Job may attempt to drive load before fullnodes can serve catch-up'd RPC. The LLD documents this as Open Dependency #1 and accepts the genesis-bootstrap case as the dominant v1 case (no real catch-up gap).
- **Hard-fork rollouts**: premature service rotation when fullnodes haven't caught up post-rollout.
- **Steady-state operations**: newly-added fullnodes serving traffic before catch-up.

### Cost of not addressing

Each consumer reimplements its own catch-up probe (`kubectl exec seid status | grep catching_up`) — exactly the bash-glue pattern the controller boundary exists to absorb. This issue moves the responsibility into the right place once.

## Relevant experts

- `kubernetes-specialist` — readiness probe contract, sidecar `/health` endpoint design
- `blockchain-developer` — `seid status` semantics, catching_up signal interpretation

## Proposed approach

Per the ValidationRun LLD's recommendation (option b in Open Dependency #1):

**Extend the `seictl` sidecar's HTTP `/health` endpoint to return 503 while `seid status.SyncInfo.catching_up=true`.** kubelet readiness probe consumes `/health`; Pod isn't Ready until caught up; SND `phase=Ready` automatically requires catch-up via the existing `ConditionNodesReady` chain.

Concretely:
- Sidecar `/health` handler: poll `seid status` (or its more efficient ABCI-direct equivalent), return 503 if `catching_up=true`, 200 otherwise.
- StatefulSet pod template: `readinessProbe.httpGet.path=/health` (already wired for the sidecar in most deployments — verify).
[...]

PR #143, opened 2026-04-29
docs: LLD for ValidationRun CRD (for #139)

bdchatham:
## Summary

Captures the council-driven low-level design for **ValidationRun** — a new namespaced CRD (`validation.sei.io/v1alpha1`) that orchestrates ephemeral-chain validation workloads on Harbor.

- **One CRD**, three reserved kinds — `ValidationRun` (implemented), `ValidationSuite` and `ValidationSchedule` (CRDs ship with v1, no reconcilers; reserved so future kinds don't paint v1 into a corner).
- **Discriminator union** at `spec.type`: `LoadTest` (v1) / `SequenceTest` / `IntegrationTest` (future). Avoids the Testkube `TestWorkflow` consolidation lesson.
- **Chain abstraction** at `spec.chain.{validators,fullNodes}` embeds `SeiNodeDeploymentSpec` directly (both required) — the controller materializes both fleets as owned children with chainId/role-discriminator/peer-selector injection. Cascade delete via OwnerReferences.
- **Validation rules** are `alert` + `query` against Prometheus, continuously polled per `runProperties.interval` (default 30s). One Prometheus client serves both types — `alert` rules use the synthetic `ALERTS` series. **Observability-as-test-oracle**: verdict = `workload_exit_code AND ⋀rules`.
- **`monitor-run` plan task** is the central refinement: a single async polling loop combining workload-monitoring + per-rule evaluation in each iteration. `Conditions[TestComplete]` is the workload-completion boundary; `runProperties.stopOnFailure` enables cooperative Job cancel on rule trip.
- **Same-namespace by construction** with one narrow exception (`alert.ruleRef` into label-allowlisted monitoring namespaces). Tenants pre-provision ServiceAccounts; the controller is never an IAM controller.
- **Workload contract parity** with sei-protocol/platform#235 — env vars, exit codes (`0/1/2`), `${RESULT_DIR}` artifacts, `.status.report.s3Url` only (no termination-message echo into status).
- **Plan integrates with existing `internal/planner/`** — same Build → Persist → Execute → Complete/Fail lifecycle, single-patch model, condition ownership on planner.

## Council process

This LLD is the artifact of a `/council` Component-tier design pass. Process documented in the local checkpoint (not committed; lives at `.council/workstream.yaml`).

- **Round 1** (parallel): product-manager (scope cuts), opentelemetry-expert (rule semantics + controller observability), platform-engineer (Harbor integration / RBAC / IRSA / GitOps)
[...]

PR #148, opened 2026-04-30
fix: reject SeiNodeDeployment with genesis on non-validator-role at admission (closes #145)

### LINKED ISSUE #145: Admission validation: reject SeiNodeDeployment with genesis on full-node-role deployments
## Problem

`SeiNodeDeployment.spec.genesis` is meaningful only for validator-role deployments (which run the genesis ceremony). Setting `genesis` on a full-node-role SND creates ambiguous behavior — full nodes inherit genesis from the validator ceremony's S3 artifact, so a user-provided `genesis` spec is silently ignored at runtime. Today the SND controller relies on the user not providing it.

ValidationRun's webhook (per the LLD at PR #143) rejects `chain.fullNodes.genesis` at admission via CEL, but the SND controller itself does not enforce this rule — operators applying SeiNodeDeployment directly can mis-configure without admission feedback.

## Impact

Low-frequency operator-error case, but the cleanest fix is to mirror ValidationRun's CEL rule at the SND-side admission layer so the check lives once for all consumers (ValidationRun-materialized SNDs + direct-`kubectl-apply` operators).

### Primary use case
Any operator authoring a SeiNodeDeployment for a full-node fleet who copy-pastes from a validator template and leaves the `genesis` block in. Today this silently succeeds at admission and the field is ignored at runtime — confusing.

### Cost of not addressing
ValidationRun's CEL rule covers ValidationRun-materialized SNDs only. Direct SND operators don't get the check. Two paths to enforce one invariant.

## Relevant experts

- `kubernetes-specialist` — CEL admission rule design

## Proposed approach

Add a CEL `XValidation` rule on `SeiNodeDeploymentSpec` rejecting `genesis` when the deployment's role is `fullNode`. Mirror ValidationRun's existing CEL rule:

```go
// +kubebuilder:validation:XValidation:rule="!has(self.genesis) || (has(self.template) && has(self.template.spec) && has(self.template.spec.validator))",message="genesis is meaningful only for validator-role deployments; remove genesis or set template.spec.validator: {}"
```

(Exact predicate may differ depending on how role is expressed in the SND spec — adjust at implementation time. The intent is "genesis present implies role=validator.")

## Acceptance criteria

- [ ] CEL XValidation rule on `SeiNodeDeploymentSpec` rejecting `genesis` on non-validator-role SNDs
- [ ] Admission rejects with a clear message naming the field and the fix
- [ ] Existing valid SeiNodeDeployment manifests in `manifests/samples/seinodedeployment/` continue to admit
- [ ] Update CRD docs to document the constraint

## Out of scope

- Migration of existing on-cluster SNDs that may have stale `genesis` blocks (admission rule applies to new/updated objects; existing stored objects are unaffected until next update)
- Changes to the role discriminator field shape

## References

- ValidationRun LLD: sei-protocol/sei-k8s-controller#143 — Open Dependency #3
- ValidationRun's CEL rule: `chain.fullNodes.genesis` rejection (search the LLD for the rule text)

bdchatham:
## Summary

Adds a CEL `XValidation` rule on `SeiNodeDeploymentSpec` rejecting `genesis` when `template.spec.validator` is unset (i.e., the deployment is full-node / archive / replayer-role, or has no role discriminator yet).

Previously, the controller silently dropped a user-supplied genesis spec on non-validator-role deployments — `internal/controller/nodedeployment/nodes.go` only stamps `GenesisCeremonyNodeConfig` when `spec.Validator != nil`, so genesis was ignored at runtime with no admission feedback. Operators who copy-pasted from a validator template into a full-node manifest got silent misconfiguration; now admission rejects with a clear message naming the field and the fix.

This is the SND-side analog of ValidationRun's webhook-level rejection of `chain.fullNodes.genesis` (per LLD merged in PR #143). Putting the rule at the SND admission boundary covers both ValidationRun-materialized SNDs and direct-`kubectl-apply` operators in one place.

## Predicate choice — Option A

Considered two predicate shapes:
- **A. `!has(self.genesis) || has(self.template.spec.validator)`** *(landed)* — rejects genesis on full-node, archive, replayer, missing-discriminator. SND has a 4-role discriminator; this matches the SND-side framing of \"genesis is meaningful only for validator-role.\"
- B. `!has(self.genesis) || !has(self.template.spec.fullNode)` — rejects only the `fullNode` case; would mirror ValidationRun's CEL exactly but ValidationRun's `role` enum knows only `validator|fullNode`, leaving archive/replayer in the silent-drop trap at the SND boundary.

A is strictly stronger and aligns with the runtime behavior — `nodedeployment/nodes.go` only stamps genesis when `spec.Validator != nil` regardless of which other discriminator is set.

## Test plan

- [x] `make manifests generate` — regenerates CRD YAML; verified the new XValidation entry appears in both `config/crd/sei.io_seinodedeployments.yaml` and `manifests/sei.io_seinodedeployments.yaml`
- [x] `make lint` — 0 issues
- [x] `make test` — all packages pass
- [x] Existing samples verified to admit unchanged: `genesis-ceremony.yaml` (validator + genesis ✓), `fork-genesis-ceremony.yaml` (validator + genesis with fork ✓), `pacific-1-rpc-group.yaml` (fullNode, no genesis ✓)

### Deferred — admission test infra

The repo has no existing CEL admission tests under `api/v1alpha1/` (the four other `XValidation` rules in the file rely on the generated CRD YAML diff as the contract). Standing up envtest + admission-test scaffolding to cover this one rule would mix scope; following the repo convention here. If the project wants admission tests broadly, that's a follow-up that should backfill all five existing `XValidation` rules at once.

## Files changed

| File | LoC | Source/generated |
|---|---|---|
| `api/v1alpha1/seinodedeployment_types.go` | +2 | source |
| `manifests/sei.io_seinodedeployments.yaml` | +5 | generated |
| `config/crd/sei.io_seinodedeployments.yaml` | +5 | generated |

## References

- Closes #145
- Related: PR #143 (ValidationRun LLD), companion issues #144 #146 #147

### DIFF
--- a/api/v1alpha1/seinodedeployment_types.go
+++ b/api/v1alpha1/seinodedeployment_types.go
@@ -5,6 +5,8 @@ import (
 )
 
 // SeiNodeDeploymentSpec defines the desired state of a SeiNodeDeployment.
+//
+// +kubebuilder:validation:XValidation:rule="!has(self.genesis) || has(self.template.spec.validator)",message="genesis is meaningful only for validator-role deployments (full nodes inherit genesis from the validator ceremony's S3 artifact); remove spec.genesis or set template.spec.validator: {}"
 type SeiNodeDeploymentSpec struct {
 	// Replicas is the number of SeiNode instances to create.
 	// +kubebuilder:validation:Minimum=1
--- a/config/crd/sei.io_seinodedeployments.yaml
+++ b/config/crd/sei.io_seinodedeployments.yaml
@@ -862,6 +862,11 @@ spec:
             - template
             - updateStrategy
             type: object
+            x-kubernetes-validations:
+            - message: 'genesis is meaningful only for validator-role deployments
+                (full nodes inherit genesis from the validator ceremony''s S3 artifact);
+                remove spec.genesis or set template.spec.validator: {}'
+              rule: '!has(self.genesis) || has(self.template.spec.validator)'
           status:
             description: SeiNodeDeploymentStatus defines the observed state of a SeiNodeDeployment.
             properties:
--- a/manifests/sei.io_seinodedeployments.yaml
+++ b/manifests/sei.io_seinodedeployments.yaml
@@ -862,6 +862,11 @@ spec:
             - template
             - updateStrategy
             type: object
+            x-kubernetes-validations:
+            - message: 'genesis is meaningful only for validator-role deployments
+                (full nodes inherit genesis from the validator ceremony''s S3 artifact);
+                remove spec.genesis or set template.spec.validator: {}'
+              rule: '!has(self.genesis) || has(self.template.spec.validator)'
           status:
             description: SeiNodeDeploymentStatus defines the observed state of a SeiNodeDeployment.
             properties:


bdchatham:
## Review (kubernetes-specialist)

### Suggestions only

Predicate is correct, idiomatic, and matches the runtime behavior in `internal/controller/nodedeployment/nodes.go`. Generated diff is minimal and clean. Approving the substance with a few small notes — none block merge.

### Findings

**Predicate correctness**
- **APPROVED — predicate handles the empty-template case cleanly.** `SeiNodeDeploymentSpec.Template` and `SeiNodeTemplate.Spec` are both **required** (no `+optional`, no `omitempty` on `Spec`), so the `required:` keyword in the generated schema rejects missing `template`/`template.spec` *before* CEL runs. `has(self.template.spec.validator)` therefore evaluates against a guaranteed-present `template.spec`. No need to defensively chain `has(self.template) && has(self.template.spec) && ...` like issue #145's draft suggested. Predicate as written is the minimal correct form.
- **APPROVED — over-rejection analysis.** The rule rejects `genesis` on `fullNode`, `archive`, `replayer`, and the no-discriminator case. The no-discriminator case is *also* caught by the existing `SeiNodeSpec` exactly-one rule, so the only operator-visible additional rejections are full-node / archive / replayer with genesis — all of which are silent-drops today per `nodes.go:233` (`gc != nil && spec.Validator != nil`). Strictly correct.
- **APPROVED — Option A over Option B.** PR body's framing is right: SND has 4 roles, ValidationRun's role enum has 2. Mirroring ValidationRun's CEL (`!has(self.template.spec.fullNode)`) at the SND boundary would leave `archive` and `replayer` in the silent-drop trap. Option A matches the runtime invariant exactly.
- **NIT — predicate doesn't catch `genesis` set on a `validator{}` deployment that is *also* somehow malformed.** Not this PR's job; the existing `SeiNodeSpec` exactly-one rule covers it.

**Idiomatic style**
- **APPROVED — placement and marker form match the four other `XValidation` markers** (`UpdateStrategy`, `SeiNodeSpec` exactly-one, `SeiNodeSpec` replayer-peers, `DataVolumeSpec` import-immutable, `ValidatorSpec` signing-key invariants). Above-type placement, double-quoted rule, single-quoted message, no line wrapping — consistent with the file.
- **NIT — message phrasing.** Slightly long compared to the other rules in this file (e.g., `"hardFork strategy requires haltHeight > 0"`, `"peers is required when replayer mode is set"`). The parenthetical (`full nodes inherit genesis from the validator ceremony's S3 artifact`) is useful operator context but pushes the message past the others' brevity bar. I'd keep it — operator-facing admission errors benefit from naming the fix path. Worth flagging only because it's the longest message in the file.

**Generated-artifact correctness**
- **APPROVED.** Diff in `config/crd/sei.io_seinodedeployments.yaml` and `manifests/sei.io_seinodedeployments.yaml` is the expected `+5 lines` `x-kubernetes-validations` entry under the spec object schema, identical between the two files. No spurious churn (timestamp drift, ordering changes, schema-id bumps). `make manifests generate` produced exactly the delta the source change implies.

**Runtime-behavior framing verification**
- **APPROVED — PR body's claim is accurate.** `internal/controller/nodedeployment/nodes.go:233` reads `if gc := group.Spec.Genesis; gc != nil && spec.Validator != nil { ... GenesisCeremony = ... }`. Genesis is silently dropped from `Validator.GenesisCeremony` stamping when `spec.Validator == nil`. Admission rejection is the right surface for this — saves the operator a debugging round-trip into "why did my full-node SND ignore my genesis spec."

**Sample admission**
- **APPROVED — verified all three samples admit.**
  - `genesis-ceremony.yaml`: `genesis: {chainId: ...}` + `template.spec.validator: {}` → `has(self.genesis) && has(self.template.spec.validator)` → predicate true. Admits.
  - `fork-genesis-ceremony.yaml`: `genesis: {chainId, fork: ...}` + `template.spec.validator: {}` → admits.
  - `pacific-1-rpc-group.yaml`: no `genesis`, `template.spec.fullNode: {snapshot: ...}` → `!has(self.genesis)` short-circuits → admits.

**Test-infra deferral**
- **APPROVED.** Repo has zero envtest-based admission tests for the four pre-existing `XValidation` rules; standing up envtest scaffolding to cover this one rule mixes scope. Following convention is correct here. **SHOULD fix (separate issue, not this PR):** file a follow-up to backfill admission tests for all five `XValidation` rules at once when envtest infra lands. The five rules (exactly-one mode, replayer-peers, signing-key/node-key Secret separation, hardFork-haltHeight, this one) are exactly the kind of invariants where regression risk is non-zero and unit-test coverage is cheap once the harness exists.

**Other**
- **NIT — docs update from issue #145's acceptance criteria.** Issue lists "Update CRD docs to document the constraint." I don't see a docs change in the diff. If there's a generated CRD reference somewhere (`docs/api-reference.md` or similar), it may pick this up automatically from the marker; if not, worth a one-line addition. Not a blocker — the message itself is self-documenting at admission time.
- **APPROVED — no security/multi-tenancy concerns.** CEL admission rules run at the apiserver, no controller-side bypass surface. No RBAC change needed.
- **APPROVED — future-compat.** When `archive`/`replayer` roles eventually grow their own ceremony semantics (unlikely but possible), this rule would need revisiting. For v1alpha1 it's correct.

### Counter-suggestions

None substantive. If you want to tighten the message to match the file's brevity convention, candidate:

```
"genesis is meaningful only for validator-role deployments; remove spec.genesis or set template.spec.validator: {}"
```

— drops the parenthetical. Operator gets the field name and the fix path; the "why" is recoverable from the LLD. I'd leave the current phrasing as-is though; the parenthetical is doing real work for operators who don't read LLDs.

### Files inspected

- `/Users/brandon/sei-k8s-controller/api/v1alpha1/seinodedeployment_types.go`
- `/Users/brandon/sei-k8s-controller/api/v1alpha1/seinode_types.go`
- `/Users/brandon/sei-k8s-controller/api/v1alpha1/common_types.go`
- `/Users/brandon/sei-k8s-controller/api/v1alpha1/validator_types.go`
- `/Users/brandon/sei-k8s-controller/internal/controller/nodedeployment/nodes.go`
- `/Users/brandon/sei-k8s-controller/manifests/samples/seinodedeployment/genesis-ceremony.yaml`
- `/Users/brandon/sei-k8s-controller/manifests/samples/seinodedeployment/fork-genesis-ceremony.yaml`
- `/Users/brandon/sei-k8s-controller/manifests/samples/seinodedeployment/pacific-1-rpc-group.yaml`
- `/Users/brandon/sei-k8s-controller/config/crd/sei.io_seinodedeployments.yaml` (PR diff)
- `/Users/brandon/sei-k8s-controller/manifests/sei.io_seinodedeployments.yaml` (PR diff)
- `/Users/brandon/sei-k8s-controller/docs/design/validation-run-lld.md` (lines 227, 251-252, 1389, 1508 — LLD framing)
- `/Users/brandon/sei-k8s-controller/CLAUDE.md`

bdchatham:
Re: the Coral reviewer's NIT on the docs acceptance criterion — confirmed the repo has no separate CRD-reference docs (the design docs at `docs/design/` are LLDs, not CRD references). CRD documentation in this repo is auto-generated from kubebuilder marker comments + the validation rule's `message` field. Both ship via this PR; the constraint is documented in the rejection message that operators see at `kubectl apply` time.

PR #157, opened 2026-05-01
feat(planner): force seid logging.level=info in config-apply overrides

### LINKED ISSUE #155: feat(controller): enforce log_level=info in seid config render
## Problem

Nothing prevents a SeiNodeDeployment spec (or downstream config rendering) from setting `log_level = debug` in seid's `config.toml`. DEBUG mode would 10-50x log volume per round-2 measurements (every `tendermint/internal/p2p/*` gossip message, every consensus block-part). At Sei's projected fleet size this would saturate Loki's ingest tier within minutes and burn cardinality budget irreversibly.

### Future-proofing

A single misconfig (CRD spec, kustomize patch, debugging session left in production) becomes a denial-of-service against the entire logging pipeline. The volume-spike alert (separate issue) provides detection, but enforcement at the source is the right gate.

## Architectural constraints

1. **Controller renders `config.toml`** from CRD spec. This is the right enforcement point.
2. **No CRD field for log_level today.** Don't add one — enforcement means refusing the override path entirely.
3. **Companion safety net.** `LokiVolumeSpike` alert (separate issue) provides detection if enforcement is somehow bypassed.

## Proposed approach

In sei-k8s-controller's seid `config.toml` render path:

1. Hard-set `log_level: info` after merging user-supplied config. User overrides for this key are dropped (with a controller log line noting the override was rejected, for debugging visibility).
2. Add a unit test asserting the rendered config has `log_level: info` regardless of input.
3. If the seid `config.toml` rendering already lives in the sidecar's `config-apply` task, enforce there; otherwise enforce in the controller's plan task that builds the final config.

**This proposal locks the following backward-compat surfaces:**
- `log_level` enum (info | warn | error). Future need for debug requires explicit escape hatch (e.g., a debug-bound CRD flag with mandatory short TTL).

## Acceptance criteria

- [ ] Test: deploy SeiNode with attempted `log_level=debug` in any input path → final config.toml shows `info`
- [ ] Existing SeiNodes pick up `info` on next reconcile (or document if it requires StatefulSet bounce)
- [ ] Controller emits a log line when an override is rejected (for ops visibility)

## Out of scope

- Per-validator debug escape hatch (file as separate issue if/when needed)
- Existing SeiNodes already running with debug (none observed)

## References

- Design: `docs/designs/loki-strategy.md` §Controller-side: enforce log_level=info
- Round-2 measurements: validator ~600 MiB/day at INFO; DEBUG would scale 10-50x

bdchatham:
## Summary

Closes #155. Locks `logging.level=info` in the override map that the controller passes to the sidecar's config-apply task. User-supplied `Spec.Overrides` cannot escape this — `logging.level=debug` at projected validator/RPC volume would 10-50x the Loki ingest rate per round-2 measurements, and could DoS the logging pipeline before any volume-spike alert ([sei-protocol/platform docs/designs/loki-strategy.md](https://github.com/sei-protocol/platform/blob/main/docs/designs/loki-strategy.md)).

## Mechanism

`mergeOverrides` already merges `controllerOverrides + userOverrides` (user wins). New `applyForcedOverrides` runs after the merge and stamps the forced key, logging a rejection line if the user supplied a different value.

```go
const (
    overrideKeyLoggingLevel = "logging.level"
    enforcedLoggingLevel    = "info"
)

func applyForcedOverrides(merged, userOverrides map[string]string) {
    if got, ok := userOverrides[overrideKeyLoggingLevel]; ok && got != enforcedLoggingLevel {
        ctrl.Log.WithName("planner").Info(
            "rejecting user override",
            "key", overrideKeyLoggingLevel,
            "user_value", got,
            "forced_value", enforcedLoggingLevel,
        )
    }
    merged[overrideKeyLoggingLevel] = enforcedLoggingLevel
}
```

Single insertion point covers every role planner (validator, full, archive, replayer) and the bootstrap genesis path — they all flow through `mergeOverrides`. Nested merge calls converge correctly: inner forces info, outer accepts user input that may overwrite it, then re-forces info (with the rejection log).

## One-way doors

This proposal locks the following backward-compat surfaces:
- `logging.level` value on every controller-managed SeiNode is now `info`. Per-node debug escape is explicitly out of scope (issue #155); future need files a separate issue with a TTL-bound CRD field.

## Test plan

- [x] `TestMergeOverrides_ForcesLoggingLevelInfo` — table test covering user_debug / user_warn / user_info / user_empty
- [x] `TestMergeOverrides_PreservesNonForcedUserKeys` — non-forced user overrides still take precedence
- [x] `TestMergeOverrides_NestedCallsConvergeToInfo` — nested merge with debug input still ends at info
- [x] Existing `TestCommonOverrides_*` continue to pass
- [x] `gofmt -s -l` clean
- [x] Existing SeiNodes pick up `info` on next reconcile (no StatefulSet bounce — sidecar config-apply is a hot path, runs without pod restart)

## Notes

- The signature accepts `userOverrides` as the rejection-detection source. In the nested call site (e.g. fullNode/archive), the inner mergeOverrides treats `controllerOverrides()` as `userOverrides`. Since none of the role planners set `logging.level`, no spurious rejection log fires there. Outer call passes `Spec.Overrides`, which is where user input arrives.
- `mergeOverrides` previously returned nil when both inputs were empty; now always returns a map containing at least `logging.level=info`. Callers pass the result to `task.ConfigApplyParams.Overrides` (`omitempty`); the field will now always be present (with at minimum the forced key). Sidecar already handles populated override maps; no compat issue.

### DIFF
--- a/internal/planner/planner.go
+++ b/internal/planner/planner.go
@@ -44,6 +44,11 @@ const (
 	TaskAwaitNodesRunning      = task.TaskTypeAwaitNodesRunning
 )
 
+const (
+	overrideKeyLoggingLevel = "logging.level"
+	enforcedLoggingLevel    = "info"
+)
+
 // baseProgression defines the ordered task sequence for each bootstrap mode.
 var baseProgression = map[string][]string{
 	"snapshot":   {TaskSnapshotRestore, TaskConfigApply, TaskConfigValidate, TaskMarkReady},
@@ -715,11 +720,6 @@ func mergeOverrides(controllerOverrides, userOverrides map[string]string) map[st
 	applyForcedOverrides(merged, userOverrides)
 	return merged
 }
-
-const (
-	overrideKeyLoggingLevel = "logging.level"
-	enforcedLoggingLevel    = "info"
-)
 
 func applyForcedOverrides(merged, userOverrides map[string]string) {
 	if got, ok := userOverrides[overrideKeyLoggingLevel]; ok && got != enforcedLoggingLevel {
--- a/internal/planner/planner.go
+++ b/internal/planner/planner.go
@@ -16,6 +16,8 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	ctrl "sigs.k8s.io/controller-runtime"
+
 	seiv1alpha1 "github.com/sei-protocol/sei-k8s-controller/api/v1alpha1"
 	"github.com/sei-protocol/sei-k8s-controller/internal/controller/observability"
 	"github.com/sei-protocol/sei-k8s-controller/internal/task"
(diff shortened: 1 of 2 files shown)


internal/planner/planner.go
```diff
@@ -707,11 +709,26 @@ func buildNodeUpdatePlan(node *seiv1alpha1.SeiNode) (*seiv1alpha1.TaskPlan, erro
...
 	maps.Copy(merged, controllerOverrides)
 	maps.Copy(merged, userOverrides)
+	applyForcedOverrides(merged, userOverrides)
 	return merged
 }
+
+const (
+	overrideKeyLoggingLevel = "logging.level"
+	enforcedLoggingLevel    = "info"
+)
```
bdchatham: Can you put this at the top of the file?

PR #179, opened 2026-05-05
feat(fork-genesis): SND-driven export sub-plan + bootstrap pkg migration

bdchatham:
## Summary

Replaces the broken `submit-export-state` (sidecar tries to exec `seid` cross-container) with an SND-driven sub-plan. Implements LLD Rev 4.1 (`docs/design-fork-genesis-export-job-lld.md`, merged at #174).

**New sub-plan**: PVC → bootstrap Job → await → export Job → await → teardown. SND owns the resources directly — no SeiNode wrapper.

**Bundled with**: bootstrap migration (`internal/task/bootstrap_*.go` → `internal/task/bootstrap/`, drops `Bootstrap` prefix on exports), framework symbol promotion (`taskBase` → `Base`, etc.), and `init()`-based task type registration so domain subpackages can self-register.

## What's in this PR

- **`internal/task/bootstrap/`** — moved 8 files; framework types qualified via `task.` import; task types registered via `init()` calling new `task.Register()`.
- **`internal/task/export/`** — new domain package:
  - `resources.go` — `GenerateJob` + `exportTriggerScript` const (the seid container's `seid export → POST /v0/tasks/upload-file → poll → exit` bash, modeled on `bootstrapWaitCommand`).
  - `fork.go` — 6 task executions: `ensure-exporter-pvc`, `apply-bootstrap-job`, `await-bootstrap-job`, `apply-export-job`, `await-export-job`, `teardown-exporter`.
  - `adapter.go` — `sndToBootstrapInputs` / `sndToExportInputs` build typed inputs at exec time from `group.Spec` + `cfg.Platform`.
- **Planner** (`internal/planner/group.go`) — emits the 6-task sub-plan when `Spec.Genesis.Fork != nil`.
- **API** — `Status.Fork.ExportJobRef`. Generated DeepCopy + CRD YAML refreshed.
- **Deleted** — `internal/task/fork_export.go` (4 old task types) + its test.

## Differences from the bootstrap Job (export §A5 in the LLD)

- 600s `TerminationGracePeriodSeconds` (vs 120s) so cascade-GC mid-upload finishes the in-flight S3 multipart.
- Sidecar env omits `SEI_SNAPSHOT_*` — only the genesis bucket is in scope.
- No `ForbiddenSecretNames` invariant — no validator on the SND fork path.
- Seid command is the export+trigger+poll script, not the bootstrap halt-height script.

## Stats

35 files changed, **+453 / −1140** (net negative because `fork_export.go` is 362 LoC deleted).

## Cross-PR contract anchor

`TestExportTriggerScript_PinsContract` asserts the bash const's wire shape against seictl's `UploadFileRequest` JSON tags (`file`/`bucket`/`key`/`region`), the sidecar URL path (`/v0/tasks/upload-file`), and the status case-arms (`completed`/`failed`). PR 1's `TestUploadFileTask_ContractKeys` is the matching anchor on the seictl side.

## Not in this PR (deferred)

- **Sidecar image digest bump** in `internal/platform/platform.go` to a build that includes seictl#136. The export Job won't function in production until that's deployed, but the controller code is correct against the upcoming sidecar contract.
- **Coverage gap**: `fork.go` task executions (apply-*, await-*, teardown) don't have direct unit tests yet — only `resources.go` (GenerateJob + bash const) is tested. Happy to fold those in a follow-up commit if you want them in this PR.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` — full suite green
- [x] `goimports -l` clean
- [x] `make manifests generate` regenerated DeepCopy + CRD
- [x] Cross-review (kubernetes-specialist + platform-engineer) — dispatching now
- [x] Sidecar image digest bump (separate follow-up)
- [x] On-cluster fork-genesis smoke test against eng-bdchatham (post-deploy)


internal/noderesource/noderesource.go
```diff
@@ -477,7 +477,7 @@ func makeResources(cpu, memory string) corev1.ResourceRequirements {
 
 // signingKeyVolumes is called only from the production StatefulSet pod-spec.
 // The bootstrap Job pod-spec must never include these volumes — see the
-// safety invariant on task.GenerateBootstrapJob.
+// safety invariant on bootstrap.GenerateJob.
```
bdchatham: stale?

internal/task/bootstrap_resources.go
```diff
@@ -62,12 +62,12 @@ type BootstrapPodInputs struct {
 	ForbiddenSecretNames []string
 }
 
-// BootstrapJobName returns the bootstrap Job name for a given resource root.
+// JobName returns the bootstrap Job name for a given resource root.
```
bdchatham: Stale?

internal/task/bootstrap_resources.go
```diff
@@ -62,12 +62,12 @@ type BootstrapPodInputs struct {
...
 func BootstrapJobName(name string) string {
 	return fmt.Sprintf("%s-bootstrap", name)
 }
 
-// BootstrapLabels returns labels for bootstrap Job resources.
```
bdchatham, line 70: Stale?

internal/task/export_resources.go
```diff
@@ -0,0 +1,327 @@
...
+    STATUS=$(echo "$RESP" | grep -oE '"status":"[^"]+"' | head -1 | cut -d'"' -f4)
+    case "$STATUS" in
+        completed) exit 0 ;;
+        failed)
+            echo "$RESP" | grep -oE '"error":"[^"]+"' | head -1 | cut -d'"' -f4 >&2
+            exit 1 ;;
+        *) sleep 5 ;;
+    esac
+done
+`
```
bdchatham, line 331: Seems like where are doing a lot of things here. Why not have these as separate tasks run by the controller?

Let's have a follow-up PR to break this up into individual tasks

Issue #230, opened 2026-05-13
Close SeiNode pod hardening gap: non-root seid + seid-init for PSA restricted eligibility


bdchatham:
## Problem

The SeiNode pod has an internally inconsistent security posture:

- `sei-sidecar` runs hardened: `RunAsUser=65532`, `ReadOnlyRootFilesystem=true`, drop ALL caps (introduced in #220 for operator-keyring containment).
- `seid` main runs as **root** — image default, no `SecurityContext` override. Cosmos SDK / CometBFT upstream convention.
- `seid-init` runs as **uid 0** with `CAP_CHOWN, CAP_FOWNER, CAP_FSETID` — explicitly added in #229 to chown the PVC and prepare shared dirs the non-root sidecar can write into.

The chown/setgid layer in #229 is a workaround for the asymmetry. It works, but it wouldn't exist if all three containers shared uid 65532. Today the pod has root surfaces (seid main + seid-init) that don't need to be there.

Surfaces:
- `internal/noderesource/noderesource.go:413` (current `PodSecurityContext` including `supplementalGroups`)
- `internal/noderesource/noderesource.go:620` (current `buildSeidInitContainer` with chown/setgid layer)
- `internal/noderesource/noderesource.go:798` (current `sidecarSecurityContext` from #220)

## Impact

While this asymmetry persists:

- Namespaces hosting SeiNodes can only enforce Pod Security Standards **`baseline`**, not `restricted` — the init container's uid 0 violates `restricted`'s `RunAsNonRoot=true` requirement.
- Cluster-wide programmatic enforcement of "no root containers" (via PSA `restricted` labels or Kyverno policies) cannot be turned on for SeiNode-hosting namespaces — every new SeiNode pod would be rejected at admission.
- The keyring-containment work in #220 is undercut conceptually: a non-root sidecar holds signing material, but a co-resident root container shares the pod's PID/IPC/network namespaces. Sidecar hardening is real but not airtight.
- Future security-hardening initiatives (Kyverno admission policies, signed-digest enforcement, allowed-registry pinning) are gated on first removing the root surface.

## Relevant experts

- `kubernetes-specialist` — SecurityContext + PVC migration init container design
- `security-specialist` — adversarial review, eventual Kyverno policy authoring
- `platform-engineer` — rollout cadence, namespace PSA label flips, cluster-side audit

## Proposed approach

Sketch, not the design. A `/coral` or `/council` pass on pickup should refine.

[...]

PR #220, opened 2026-05-12
feat: SeiNode validator.operatorKeyring CRD surface (closes #219)

bdchatham:
## Summary

- Implements Phase 1 step 1 of the in-pod governance signing workstream (design: sei-protocol/seictl#166 / [`docs/design/in-pod-governance-signing.md`](https://github.com/sei-protocol/seictl/blob/main/docs/design/in-pod-governance-signing.md) — Component A).
- Adds `validator.operatorKeyring` to `SeiNode` — discriminated-union surface (`Secret` variant; future `TMKMS`/`Vault`/`KMS` reserved as comments and an additive CEL `exactly-one` rule) declaring the operator-account keyring used by the seictl sidecar for governance, `MsgEditValidator`, withdraw-rewards, and other operator-account transactions.
- **Trust boundary held at admission, build-time, AND runtime**: four CEL invariants on `ValidatorSpec`; pod-spec mount + env wired into the **sidecar container only** (never seid main, never init, never bootstrap); bootstrap-pod isolation guard generalized to walk both Secret volumes AND `valueFrom.secretKeyRef` env entries.
- Sidecar container security context tightened in the same workstream (the keyring mount is precisely the reason): `RunAsNonRoot`, `RunAsUser/Group 65532`, `ReadOnlyRootFilesystem`, `AllowPrivilegeEscalation: false`, `Capabilities.Drop: [ALL]`, `SeccompProfile: RuntimeDefault`, pod-level `FSGroup: 65532` (required for the non-root UID to read the 0o400 Secret mount).
- Pre-flight `validate-operator-keyring` task mirrors `validate-signing-key.go`. Deliberately does NOT decrypt the keyring — that's the sidecar's startup smoke test (issue #162 in seictl), not the controller's TCB.

## CRD invariants

Four CEL rules on `ValidatorSpec` enforce the trust boundary at admission:
1. `operatorKeyring.secret.secretName != signingKey.secret.secretName` (data-keyring ≠ consensus-key Secret)
2. `operatorKeyring.secret.secretName != nodeKey.secret.secretName` (data-keyring ≠ P2P-key Secret)
3. `operatorKeyring.secret.secretName != passphraseSecretRef.secretName` (data-keyring ≠ passphrase Secret)
4. `secretName` and `passphraseSecretRef.secretName` immutable via `self == oldSelf`

`keyName` is intentionally **NOT** immutable (allows key rotation within a Secret).

## Discretionary decisions flagged for review

[...]

PR #232, opened 2026-05-13
feat(noderesource): migrate seid main to non-root, recursive PVC chown

bdchatham:
## Summary

Collapses the asymmetry workaround layer from #229 (and supersedes the now-closed #231) into the proper non-root seid migration that the long-term tracking issue #230 called for.

- **seid main** now runs as `RunAsUser=65532, RunAsGroup=65532, RunAsNonRoot=true` — matches the sidecar's uid/gid from #220.
- **seid-init script** extracted to `seid_init.sh.tmpl` + `text/template` render. Adds a sentinel-gated one-time `chown -R 65532:65532 /sei` after `seid init`, gated on `/sei/.sei-k8s-controller-ownership-v1`. Sentinel const exported (`OwnershipSentinelV1`) so bootstrap_resources.go can clear it on re-bootstrap.
- **bootstrap Job's seid-init** now `rm -f`s the sentinel before exit (re-bootstrap on a previously-migrated PVC otherwise writes root-owned files past the stamped sentinel; seid main would EACCES).
- **`PodSecurityContext.SupplementalGroups: [65532]`** removed — the workaround that gave root-seid group access to gid-65532-owned files is no longer needed once seid IS gid 65532.

## Why this shape

Three trio rounds. The shape that won out:

1. **Symmetric uid/gid** between seid main and sei-sidecar is the load-bearing simplification. Owner-uid match shortcircuits the kernel permission check in `generic_permission()`; group/setgid/supplementalGroups workarounds disappear.
2. **One-time PVC migration via sentinel** rather than recursive-chown-every-restart amortizes the cost. First-restart per PVC pays the chown; subsequent restarts skip via single `stat`.
3. **seid-init stays at uid 0** because `chown(2)` needs `CAP_CHOWN`. Drops every other cap; keeps `CHOWN/FOWNER/FSETID` only.
4. **Template extraction**: the inline 13-arg `fmt.Sprintf` was unwieldy. Now the script lives in `seid_init.sh.tmpl` with named template fields; the Go side just hands it `{DataDir, NonRootUID, SentinelName, ChainID}`.

## Rollback safety

Rolling back to a pre-this-PR controller is safe: the old controller renders seid as uid 0. uid 0 has `CAP_DAC_OVERRIDE` and bypasses DAC checks against 65532-owned files. No filesystem migration is needed to roll back. Worth noting: new files seid (uid 0) writes during a rollback window are owned 0:0; if you roll forward again, the sentinel still exists, so those root-owned files would lock out seid 65532. Recovery: `rm /sei/.sei-k8s-controller-ownership-v1` on the affected PVC to re-trigger migration. Document this in the rollout runbook.

## Pod Security Standards

seid-init still runs as uid 0 (chown needs it), so namespaces hosting SeiNode pods must enforce `pod-security.kubernetes.io/enforce=baseline`, **not** `restricted`. Full `restricted` eligibility is a future workstream (init must drop to uid 65532 once the fleet is fully migrated and the chown branch becomes dead code, or replace seid-init with a kubelet-driven `FSGroupChangePolicy=Always` approach).

## First-rollout operational impact

Every existing SeiNode StatefulSet drifts on this controller's rollout. The `NodeUpdate` plan's `replace-pod` task fires per-node, rolling each pod individually. **First restart per PVC pays the recursive chown cost**:
- Harbor smoke chains (small datasets): seconds.
- Prod archive nodes (~10TB / ~50M files on EBS gp3): estimated ~10-30 min one-time. Subsequent restarts: single `stat` on the sentinel, near-zero.

Existing chain data is preserved — `chown -R` is a metadata-only walk; file contents are untouched.

**`Failed`-phase SeiNodes don't auto-reconcile** onto the new pod template (the planner has no path back from terminal Failed). The two `harbor-pr-229-{0,1}` validators currently in Failed state need `kubectl delete seinode` + SND recreate to pick up the migration; harbor is small so this is acceptable as the smoke-test reset.

## Deferred to follow-up tracking issues

- **Sentinel as JSON content marker** (vs plain dotfile) before introducing migration-v2 — path-based versioning is brittle; security flagged.
- **Parallel chown for multi-TB archives** — current `chown -R` is single-threaded. `find … -print0 | xargs -0 -P N chown` would parallelize. Not needed for the harbor smoke; could matter for prod archive nodes if the 10-30 min estimate is wrong.
- **Operator recovery runbook entry**: `rm /sei/.sei-k8s-controller-ownership-v1` + restart pod when migration goes wrong on a specific PVC.
- **Migration-in-progress event emission** — emit a controller event when the init container is running pre-sentinel so operators see it in `kubectl describe`.
- **Shell-exec idempotency test harness** — current tests check rendered strings; an actual shell execution against a tmpdir would catch regressions like the Issue A bug (which was caught by code review, not tests).
- **Replace seid-init with kubelet `FSGroupChangePolicy=Always`** — security flagged as the long-long-term shape; deletes the entire init container in favor of CSI-driven recursive chown. Unblocks PSA `restricted`.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l .`
- [x] `go test ./...`
- [x] `go test -race ./internal/noderesource/...`
- [x] `golangci-lint run --new-from-rev=origin/main` (0 issues)
- [x] Coral trio (platform-engineer, kubernetes-specialist, security-specialist) — three rounds, all converged on the current shape. Real bug caught in round 2 (Issue A: trailing `mkdir -p /sei/tmp` post-sentinel) and fixed.
- [ ] Smoke: harbor controller bumped, fresh genesis chain reaches Ready end-to-end. Two old Failed SeiNodes deleted; SeiNodeDeployment recreates fresh.

### DIFF
--- a/internal/noderesource/noderesource.go
+++ b/internal/noderesource/noderesource.go
@@ -5,9 +5,12 @@
 package noderesource
 
 import (
+	"bytes"
+	_ "embed"
 	"fmt"
 	"maps"
 	"strings"
+	"text/template"
 
 	seiconfig "github.com/sei-protocol/sei-config"
 	appsv1 "k8s.io/api/apps/v1"
@@ -21,6 +24,22 @@ import (
 	seiv1alpha1 "github.com/sei-protocol/sei-k8s-controller/api/v1alpha1"
 	"github.com/sei-protocol/sei-k8s-controller/internal/platform"
 )
+
+// OwnershipSentinelV1 is the per-PVC marker for migration-v1 (data volume
+// chowned to sidecarNonRootUID:sidecarNonRootUID). Exported because
+// internal/task/bootstrap_resources.go must clear it on re-bootstrap so
+// the StatefulSet's seid-init re-runs the migration over whatever the
+// bootstrap Job rewrote as uid 0.
+//
+// Any future ownership migration MUST define a new sentinel (V2, etc.)
+// rather than reuse this path — old PVCs would otherwise short-circuit
+// the new migration on a stale marker.
+const OwnershipSentinelV1 = ".sei-k8s-controller-ownership-v1"
+
+//go:embed seid_init.sh.tmpl
+var seidInitScriptTmpl string
+
+var seidInitTemplate = template.Must(template.New("seid-init").Parse(seidInitScriptTmpl))
 
 const (
 	// NodeLabel is the standard label key used on all SeiNode-owned resources.
@@ -423,10 +442,6 @@ func buildNodePodSpec(node *seiv1alpha1.SeiNode, p PlatformConfig) corev1.PodSpe
 	spec.SecurityContext = &corev1.PodSecurityContext{
 		FSGroup:             &fsGroup,
 		FSGroupChangePolicy: &fsGroupChangePolicy,
-		// seid (uid 0) writes into gid-65532-owned dirs created by seid-init,
-		// so its supplementary groups must include the sidecar gid to pick up
-		// group-write perms on /sei/config and /sei/data.
-		SupplementalGroups: []int64{sidecarNonRootUID},
 	}
 	initContainers := []corev1.Container{
 		buildSeidInitContainer(node),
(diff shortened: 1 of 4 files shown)


bdchatham:
Scrapped in favor of the simpler shape: kubelet's fsGroup OnRootMismatch handles the chown migration natively, so the embedded shell + sentinel + bootstrap re-handling complexity is unnecessary. Coral trio (k8s + platform + security) independently converged on this verdict after online-research-validated review. Replacement PR incoming.

PR #254, opened 2026-05-16
feat(planner): TLS toggle on Running SeiNodes via NodeUpdate plan

bdchatham:
## Summary

- Extends `buildRunningPlan` to detect `spec.sidecar.tls` drift alongside image drift; both flow into one extended `buildNodeUpdatePlan(node, imageDrift, tlsDrift)` so a single pod cycle covers both.
- New `WaitForSidecarTLSSecret` task gates `ApplyStatefulSet` on cert-manager async issuance; reads `Certificate.Ready` and returns `Terminal` on `Reason=Failed` so misconfigured Issuers fail fast with `UpdateFailed`.
- New `ObserveSidecarTLS` task stamps `status.currentSidecarTLS` after StatefulSet rollout convergence; mirrors `ObserveImage` pattern.
- One `ConditionNodeUpdateInProgress` condition covers both drifts; `reason` discriminates (`UpdateStarted` / `TLSToggleStarted` / `UpdateAndTLSToggleStarted`).

Closes #247.

## Design choice: extend NodeUpdate, don't add a parallel plan

A separate `SidecarReprovision` plan + `SidecarTLSToggleInProgress` condition was considered and rejected — see LLD §0.1 in \`docs/design-seinode-sidecar-tls-toggle-lld.md\`. The kube-rbac-proxy shares a pod with seid, so the pod-cycle middle of both plans was identical. The implicit \"image takes precedence, regenerates StatefulSet with proxy\" co-drift rule would leave \`status.currentSidecarTLS\` unstamped on the next reconcile, triggering a redundant second pod cycle. Discriminating image rollout vs. TLS toggle on dashboards is delivered via \`condition.reason\` instead.

## Scope

- \`api/v1alpha1\`: \`SidecarTLSStatus\` type + \`SeiNodeStatus.CurrentSidecarTLS\` mirror; CRD additive
- \`internal/planner\`: \`sidecarTLSDrift\` (enable-only); extended \`buildNodeUpdatePlan\`; \`nodeUpdateReason\` / \`nodeUpdateMessage\` helpers; \`classifyPlan\` new labels (\`tls-toggle\`, \`node-update-tls\`)
- \`internal/task\`: \`WaitForSidecarTLSSecret\` (Cert + Secret gate with cert-manager terminal-failure path), \`ObserveSidecarTLS\` (status stamper)
- Drop the \"Init-only\" doc-comment on \`SidecarConfig.TLS\`
[...]

Issue #263, opened 2026-05-18
Harden status.currentSidecarTLSSecretName writes via ValidatingAdmissionPolicy


bdchatham:
## Problem

Surfaced in #262 round-1 cross-review (security). With the in-place TLS-enable design, \`Status.CurrentSidecarTLSSecretName\` drives the controller's sidecar client transport selection (\`SidecarURLForNode\` + \`cmd/main.go newSidecarClient\` both read it). An actor with \`seinodes/status\` patch permission could flip the controller's transport (DoS on the reconcile loop, not data exfil — sidecar still enforces its own auth).

Existing RBAC (\`manifests/role.yaml\`) scopes \`seinodes/status\` patch to the controller's own ServiceAccount, so the trust assumption is held by RBAC today. This issue tracks defense-in-depth: a ValidatingAdmissionPolicy that rejects writes to this specific status field from any identity other than the controller SA, so an accidental RBAC mistake (overly broad ClusterRole) doesn't open the surface.

## Proposed scope

- ValidatingAdmissionPolicy + ValidatingAdmissionPolicyBinding scoped to \`seinodes.sei.io/status\` subresource
- Match expression: changes to \`object.status.currentSidecarTLSSecretName\`
- Allowed users: only the controller's ServiceAccount

## Why deferred from #262

Current RBAC scope is correct; admission-policy is defense in depth, not a missing primitive. Adds K8s 1.30+ requirement.

## References

- #262 (PR for in-place TLS enable)
- \`internal/noderesource/noderesource.go:1158-1170\` (\`SidecarURLForNode\`)
- \`cmd/main.go:177-189\` (\`newSidecarClient\`)
- \`manifests/role.yaml\`

bdchatham:
Closing — directional pivot. We're removing sidecar TLS management from the controller entirely. Rationale: controller↔sidecar runs on the private K8s cluster network; TLS termination doesn't earn its keep here. kube-rbac-proxy stays for its SAR-based authz value (in `--insecure-listen-address` HTTP mode); all cert provisioning, Secret reference contract, preflight SAN validation, status mirror, and CEL immutability surface go away.

Subtraction PR incoming.

PR #258, opened 2026-05-16
feat(seinode): sidecar TLS via externally-provisioned Secret

bdchatham:
## Summary

- Implements #255 per the merged LLD at `docs/design-seinode-sidecar-tls-toggle-lld.md`.
- Externalizes cert ownership: operator/platform tooling provisions the `kubernetes.io/tls` Secret; the controller validates SANs in a steady-state preflight branch and gates init-plan creation on `ConditionSidecarTLSSecretReady`. Mirrors `SigningKeyReady` / `NodeKeyReady` / `OperatorKeyringReady` / imported-PVC patterns already in the controller.
- `spec.sidecar.tls` is immutable post-creation via CRD CEL. Toggle = delete + recreate.

Closes #255.

## Design at a glance

| Surface | Before | After |
|---|---|---|
| `SidecarTLSSpec` | `{IssuerName, IssuerKind}` | `{SecretName}` |
| Cert provisioning | Controller mints `cert-manager.io/v1 Certificate` via `ApplySidecarCert` plan task | Operator provisions `kubernetes.io/tls` Secret externally |
| Cert validation | None (cert-manager Ready was the gate) | Controller preflight: type, non-empty `tls.crt`/`tls.key`, `x509.ParseCertificate`, SAN superset check |
| Status | Nothing | `status.sidecarTLS.{secretName, requiredDNSNames}` published as machine-readable contract |
| Condition | None | `ConditionSidecarTLSSecretReady` with `TLSSecretReady` / `TLSSecretNotFound` / `TLSSecretMalformed` / `TLSSecretSANsMismatch` |
| In-place toggle | Silently ignored on Running | CEL rejects mutation; toggle requires delete + recreate |

## What's deleted

- `internal/task/apply_sidecar_cert.go` (task + deserializer registry entry)
- `GenerateSidecarCertificate` from `internal/noderesource/noderesource.go`
- `SidecarTLSSpec.IssuerName` / `IssuerKind` fields

## Operator workflow

1. Pre-compute the required DNS names from the target spec: `{name}.{ns}.svc.cluster.local` + `{name}-0.{name}.{ns}.svc.cluster.local`. (Also published in `status.sidecarTLS.requiredDNSNames` after creation for verification.)
2. Provision a `kubernetes.io/tls` Secret with those SANs (e.g., via a cert-manager `Certificate` in GitOps).
3. Create the SeiNode with `spec.sidecar.tls.secretName: <name>`.
[...]

PR #264, opened 2026-05-18
feat(seinode): in-place TLS enable on Running SeiNodes

### LINKED ISSUE #262: Enable TLS on existing Running SeiNodes via in-place plan
## Problem

#258 landed the externalized-cert + immutable-spec model: `spec.sidecar.tls` is fixed at creation, and toggling TLS on an existing SeiNode requires delete + recreate with PVC retention (LLD §5). The deferral was deliberate at ~10 nodes — but the operational pain is real per-node, and the externalized-cert foundation #258 built actually makes in-place enable much simpler than the original PR #254 attempt was.

LLD §6 explicitly flagged this as "re-open as a separate design if the fleet scale or operational model changes." This issue is that reopen.

## Why this is easier than PR #254 was

The original in-place toggle PR was rejected because:
1. Controller was minting cert-manager `Certificate` resources → coupled cert lifecycle to plan lifecycle (drift detector + observer + wait task + cert-manager-Ready terminal logic).
2. `IssuerName`/`IssuerKind` were operator-spec-supplied → security gap (#251).
3. `classifyPlan` task-list-sniffing got brittle as new plan types accumulated.

Under #258's foundation, all three of those are dissolved:
- Cert is operator-provisioned externally; controller doesn't create it.
- No Issuer fields in spec; trust-anchor selection lives entirely outside the controller's blast radius.
- The preflight already validates the cert; the plan can rely on `SidecarTLSSecretReady=True` as a precondition rather than reproducing validation logic inline.

The residual surface is a focused planner extension plus one transport-mode consistency fix.

## Proposed approach

1. **Relax CEL.** Allow only the `tls: nil → tls: {secretName: X}` transition. Still block disable (`set → nil`) and swap (`X → Y`); those are the genuinely irreversible operations that should require delete + recreate. The current rule:
   ```
   (!has(oldSelf.sidecar) || !has(oldSelf.sidecar.tls)) ? (!has(self.sidecar) || !has(self.sidecar.tls)) : (...equality...)
   ```
   becomes:
   ```
   (!has(oldSelf.sidecar) || !has(oldSelf.sidecar.tls)) ? true : (has(self.sidecar) && has(self.sidecar.tls) && self.sidecar.tls == oldSelf.sidecar.tls)
   ```
   First branch flips from "require self also has no TLS" to "accept anything" — that's the enable case. Second branch unchanged.

2. **Reintroduce `Status.CurrentSidecarTLS` mirror.** Just `{SecretName string}` now — way simpler than PR #254's `{IssuerName, IssuerKind}` struct. Source of truth for what TLS state the live pod is actually serving (vs. what spec wants).

3. **Drift detector.** Narrow trigger: `Spec.Sidecar.TLS != nil && Status.CurrentSidecarTLS == nil && ConditionSidecarTLSSecretReady == True`. Only fires when (a) operator added TLS post-creation, (b) preflight already validated the Secret. Disable / swap remain CEL-blocked so the detector never sees those cases.

4. **Plan composition.** Extend `buildNodeUpdatePlan` to compose `ApplyRBACProxyConfig → ApplyStatefulSet → ApplyService → ReplacePod → ObserveSidecarTLS → MarkReady` when TLS-enable drift fires. Image drift and TLS-enable drift co-compose in a single pod cycle (same pattern PR #254 explored, but without the cert-creation pre-tasks).

5. **`ObserveSidecarTLS` task.** Stamps `Status.CurrentSidecarTLS = {SecretName: Spec.Sidecar.TLS.SecretName}` after rollout convergence. Mirrors `ObserveImage`.

6. **Transport-mode source switch.** This is the load-bearing fix that PR #254 missed: `SidecarURLForNode` currently reads `Spec.Sidecar.TLS` to decide HTTP-vs-HTTPS. Mid-rollout, spec disagrees with the live pod, so controller→sidecar calls fail (controller tries HTTPS against an HTTP-only pod, or vice versa). Switch the function to read from `Status.CurrentSidecarTLS` so the controller's client uses whatever transport the rollout has actually landed.

7. **`classifyPlan` label.** Add `node-update-tls-enable` for plans triggered by TLS-enable drift only; `node-update` (image-only) and the co-drift case (`node-update-tls-enable+image-update`?) follow.

8. **Tests.** Drift-detector matrix, plan composition across (image-only / tls-enable-only / both), CEL allows-enable / blocks-disable / blocks-swap, transport-mode-source switch under each plan phase.

Estimated size: ~250–400 LOC.

## Acceptance criteria

- [ ] CEL relaxed; envtest confirms enable allowed, disable + swap rejected
- [ ] `Status.CurrentSidecarTLS` field + DeepCopy regen
- [ ] `sidecarTLSEnableDrift` detector in `internal/planner/planner.go`
- [ ] `ObserveSidecarTLS` task + registry entry + paramsForTaskType entry
- [ ] `buildNodeUpdatePlan` composes pre-tasks (RBAC ConfigMap) and observer when TLS drift present
- [ ] `SidecarURLForNode` reads from status, not spec — verified by a test that exercises the mid-rollout state
- [ ] `classifyPlan` emits the new label(s)
- [ ] Unit tests for drift matrix + plan composition
- [ ] envtest for CEL transitions
- [ ] End-to-end on harbor: add `spec.sidecar.tls.secretName` to a Running SeiNode with a pre-provisioned Secret; pod cycles; serves on `:8443`

## Out of scope (deferred)

- **Disable path** (`tls: set → nil`). Still requires delete + recreate. Same as #250 was — reopen if it becomes load-bearing.
- **Swap path** (`tls: A → tls: B`). Issuer-rotation use case; same reasoning as disable. Operationally rare.
- **SND-level rolling enable** orchestrated by the parent controller. Once per-SeiNode enable works, the SND can drive it by editing the template; this issue does not address SND-level orchestration.

## References

- #258 — externalized-cert foundation (merged)
- LLD §6 — explicit deferral framing
- PR #254 — original in-place toggle (closed); informs what NOT to do
- #260, #261 — separate hardening / observability follow-ups

bdchatham:
## Summary

- Operators can now add `spec.sidecar.tls.secretName` to an existing Running SeiNode; the controller composes a NodeUpdate-style plan that cycles the pod with the kube-rbac-proxy mount.
- Builds on #258's externalized-cert foundation. Disable and swap remain CEL-rejected — set-once semantics, not full mutability.
- Mid-rollout transport-mode windowing (the bug that bit closed PR #254) is closed by sourcing transport from observed `status.currentSidecarTLSSecretName` rather than spec.

Closes #262.

## Design

| Surface | Behavior |
|---|---|
| CEL on `spec.sidecar.tls` | `nil → set` allowed; `set → nil` (disable) and `A → B` (swap) rejected |
| `Status.CurrentSidecarTLSSecretName` | Mirror stamped by `ObserveSidecarTLS` after rollout convergence (`ReadyReplicas >= *Replicas`) |
| `SidecarURLForNode` + `cmd/main.go newSidecarClient` | Both key on the same observed status field — lockstep transport flip |
| `sidecarTLSEnableDrift` | Narrow detector: spec set + status empty + preflight Ready |
| `buildNodeUpdatePlan` | Composes image-drift + TLS-enable drift in one pod cycle |
| Init plans (base / bootstrap / genesis) | Inject `ObserveSidecarTLS` between StatefulSet apply and first sidecar HTTP task |
| `classifyPlan` labels | New: `tls-enable`, `node-update-tls-enable` |

LLD §0.1(b), §4, §5, §6 rewritten in this PR to reflect set-once semantics — the original LLD said immutable; un-deferred per operator validation that per-node delete+recreate is more friction than warranted.

## Cross-review

Trio (platform / kubernetes / security) ran against `77a28c1`. Three commits address findings:

- `77a28c1` — implementation
- `e28c92b` — round-1 fixups (cmd/main.go transport gate, ReadyReplicas check, classifyPlan separator, nil-guard, LLD rewrite)
- `1f33673` — comment sweep

**Blockers addressed:**
- **k8s**: `cmd/main.go newSidecarClient` was gated on spec while `SidecarURLForNode` reads status — divergence would break mid-rollout (in-place enable) and Phase-2 bootstrap (mTLS doer against HTTP bootstrap pod). Switched gating predicate to `Status.CurrentSidecarTLSSecretName != ""`.
- **security**: Status field drives transport — audit confirms `seinodes/status` patch RBAC is scoped to the controller SA only. Documented as load-bearing trust assumption in LLD §0.1(b). Defense-in-depth admission policy tracked as #263.
- **platform + security**: LLD §0.1(b)/§4/§5/§6 said this work was "explicitly rejected." Rewrote to reflect set-once semantics, transport-mode invariant, and the design-history evolution across #258 → #262.

**Non-blocker fixups applied:**
- `ObserveSidecarTLS` now requires `ReadyReplicas >= *Replicas` (k8s) — stricter than `ObserveImage` because stamping flips transport
- `classifyPlan` label `node-update+tls-enable` → `node-update-tls-enable` (security; avoids PromQL regex footgun)
- `sidecarTLSEnableDrift` has explicit nil-guard (k8s)
- `buildBasePlan` adjacent `SidecarTLSEnabled` blocks consolidated (platform)
- Bootstrap ObserveSidecarTLS placement comment clarifies Phase-2 vs Phase-5 (k8s)
- Test comment contradiction fixed (k8s)
- Added `SidecarURLForNode` tests for both mid-rollout quadrants

**Deferred (tracked):**
- #263 — ValidatingAdmissionPolicy on `seinodes/status` writes (defense in depth)
- Envtest harness for CEL transitions (future infra)
- Shared rollout-readiness primitive between ObserveImage and ObserveSidecarTLS (lift if a third observer lands)

## Test plan

- [x] `make test` — full suite green
- [x] `golangci-lint run --new-from-rev=origin/main ./...` — 0 issues
- [x] `make manifests generate` — CRD diff is additive (new field optional, CEL change is a relaxation)
- [x] Unit: `sidecarTLSEnableDrift` matrix (fires / converged / not-ready / disabled)
- [x] Unit: plan composition across (image-only, tls-only, both) on Running nodes
- [x] Unit: init-plan ObserveSidecarTLS placement (after ApplyStatefulSet, before ConfigApply)
- [x] Unit: `SidecarURLForNode` mid-rollout quadrants (spec-set/status-empty stays HTTP; status-set drives HTTPS)
- [x] Unit: `ObserveSidecarTLS` crashlooping-proxy scenario (UpdatedReplicas=1, ReadyReplicas=0 → stays Running)
- [x] Unit: `classifyPlan` new labels
- [x] Unit: condition lifecycle on TLS-enable plans (sets and clears)
- [ ] envtest: CEL allows nil→set, rejects set→nil + A→B (deferred — envtest harness not yet in repo)
- [ ] End-to-end on harbor: patch `spec.sidecar.tls.secretName` on a Running SeiNode with a pre-applied Secret; pod cycles; serves on `:8443`

## References

- #262 — tracking issue
- #258 — externalized-cert foundation (merged)
- `docs/design-seinode-sidecar-tls-toggle-lld.md` — LLD revised in this PR

### DIFF
--- a/api/v1alpha1/seinode_types.go
+++ b/api/v1alpha1/seinode_types.go
@@ -10,7 +10,7 @@ import (
 // the populated field determines the node's operating mode.
 // +kubebuilder:validation:XValidation:rule="(has(self.fullNode) ? 1 : 0) + (has(self.archive) ? 1 : 0) + (has(self.replayer) ? 1 : 0) + (has(self.validator) ? 1 : 0) == 1",message="exactly one of fullNode, archive, replayer, or validator must be set"
 // +kubebuilder:validation:XValidation:rule="!has(self.replayer) || (has(self.peers) && size(self.peers) > 0)",message="peers is required when replayer mode is set"
-// +kubebuilder:validation:XValidation:rule="(!has(oldSelf.sidecar) || !has(oldSelf.sidecar.tls)) ? (!has(self.sidecar) || !has(self.sidecar.tls)) : (has(self.sidecar) && has(self.sidecar.tls) && self.sidecar.tls == oldSelf.sidecar.tls)",message="spec.sidecar.tls is immutable; delete + recreate the SeiNode to change TLS configuration"
+// +kubebuilder:validation:XValidation:rule="(!has(oldSelf.sidecar) || !has(oldSelf.sidecar.tls)) ? true : (has(self.sidecar) && has(self.sidecar.tls) && self.sidecar.tls == oldSelf.sidecar.tls)",message="spec.sidecar.tls is set-once: enabling on an existing SeiNode is allowed, but disabling or changing requires delete + recreate"
 type SeiNodeSpec struct {
 	// ChainID of the chain this node belongs to.
 	// +kubebuilder:validation:MinLength=1
@@ -362,6 +362,13 @@ type SeiNodeStatus struct {
 	// SidecarTLS is set when spec.sidecar.tls is non-nil.
 	// +optional
 	SidecarTLS *SidecarTLSStatus `json:"sidecarTLS,omitempty"`
+
+	// CurrentSidecarTLSSecretName is the spec.sidecar.tls.secretName
+	// observed on the live pod after the rollout converged. Drives the
+	// controller's sidecar client transport-mode selection (HTTP vs
+	// HTTPS). Empty when no TLS rollout has converged yet.
+	// +optional
+	CurrentSidecarTLSSecretName string `json:"currentSidecarTLSSecretName,omitempty"`
 }
 
 // SidecarTLSStatus is the controller's view of the referenced TLS Secret.
--- a/api/v1alpha1/seinode_types.go
+++ b/api/v1alpha1/seinode_types.go
@@ -10,7 +10,7 @@ import (
 // the populated field determines the node's operating mode.
 // +kubebuilder:validation:XValidation:rule="(has(self.fullNode) ? 1 : 0) + (has(self.archive) ? 1 : 0) + (has(self.replayer) ? 1 : 0) + (has(self.validator) ? 1 : 0) == 1",message="exactly one of fullNode, archive, replayer, or validator must be set"
 // +kubebuilder:validation:XValidation:rule="!has(self.replayer) || (has(self.peers) && size(self.peers) > 0)",message="peers is required when replayer mode is set"
-// +kubebuilder:validation:XValidation:rule="(!has(oldSelf.sidecar) || !has(oldSelf.sidecar.tls)) ? true : (has(self.sidecar) && has(self.sidecar.tls) && self.sidecar.tls == oldSelf.sidecar.tls)",message="spec.sidecar.tls is set-once: enabling on an existing SeiNode is allowed, but disabling or changing requires delete + recreate"
+// +kubebuilder:validation:XValidation:rule="(!has(oldSelf.sidecar) || !has(oldSelf.sidecar.tls)) ? true : (has(self.sidecar) && has(self.sidecar.tls) && self.sidecar.tls == oldSelf.sidecar.tls)",message="spec.sidecar.tls is set-once; disable and swap require delete + recreate"
 type SeiNodeSpec struct {
 	// ChainID of the chain this node belongs to.
 	// +kubebuilder:validation:MinLength=1
@@ -363,10 +363,8 @@ type SeiNodeStatus struct {
 	// +optional
 	SidecarTLS *SidecarTLSStatus `json:"sidecarTLS,omitempty"`
 
-	// CurrentSidecarTLSSecretName is the spec.sidecar.tls.secretName
-	// observed on the live pod after the rollout converged. Drives the
-	// controller's sidecar client transport-mode selection (HTTP vs
-	// HTTPS). Empty when no TLS rollout has converged yet.
+	// CurrentSidecarTLSSecretName is the secretName observed on the
+	// live pod. Empty when TLS hasn't rolled out yet.
 	// +optional
 	CurrentSidecarTLSSecretName string `json:"currentSidecarTLSSecretName,omitempty"`
 }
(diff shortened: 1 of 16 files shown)


bdchatham:
Closing — directional pivot. We're removing sidecar TLS management from the controller entirely. Rationale: controller↔sidecar runs on the private K8s cluster network; TLS termination doesn't earn its keep here. kube-rbac-proxy stays for its SAR-based authz value (in `--insecure-listen-address` HTTP mode); all cert provisioning, Secret reference contract, preflight SAN validation, status mirror, and CEL immutability surface go away.

Subtraction PR incoming.

Issue #266, opened 2026-05-18
Audit /v0/metrics content; now unauthenticated cluster-wide


bdchatham:
## Problem

Surfaced by security cross-review on the TLS-removal refactor. `kube-rbac-proxy --ignore-paths` exposes the sidecar's `/v0/healthz`, `/v0/livez`, `/v0/startupz`, and `/v0/metrics` without TokenReview or SAR. With the refactor making the proxy always-on and reachable from any pod in the cluster (until #266 lands), `/v0/metrics` is a passive disclosure channel.

The other three are health endpoints with known-bounded response bodies. `/v0/metrics` may export validator pubkeys, peer IDs, signing latencies, or any data correlated with operator-keyring use.

## Proposed scope

1. Enumerate exactly what the sidecar emits on `/v0/metrics` (sei-sidecar source).
2. If any field is sensitive (validator identity, signing-latency, peer-identity beyond what's already public on chain), either gate behind authz (`/v0/metrics-internal`) or move to a non-bypassed path.

## Why deferred from refactor/seinode-drop-sidecar-tls

The bypass-paths set is unchanged by the refactor; the audit applies whether or not we did the TLS removal. Easier as a focused follow-up against the sei-sidecar codebase.

## References

- `refactor/seinode-drop-sidecar-tls`
- `internal/noderesource/noderesource.go` `bypassPaths()`

Issue #265, opened 2026-05-18
NetworkPolicy: restrict :8443 ingress to the controller SA


bdchatham:
## Problem

Surfaced by security cross-review on the TLS-removal refactor. With kube-rbac-proxy running in `--insecure-listen-address` (HTTP) and SAR-based authz as the trust boundary, the controller→sidecar bearer token traverses the cluster network in clear text. Authz still works, but the threat model now leans entirely on "no packet-capture-capable workload in the cluster" and "no untrusted workload can reach :8443 on a SeiNode pod."

A per-SeiNode NetworkPolicy restricting ingress on :8443 to the controller's pod labels closes the second leg of that assumption.

## Proposed scope

- Generate a NetworkPolicy per SeiNode (or a shared one keyed on the SeiNode-owned pod labels) selecting pods with `sei.io/node` label
- `policyTypes: [Ingress]`, single rule allowing `from.podSelector` matching the controller manager pod (or `from.namespaceSelector` for the controller namespace)
- Owner-ref'd to the SeiNode so it GCs cleanly

## Why deferred from refactor/seinode-drop-sidecar-tls

The refactor establishes the new trust model; NetworkPolicy hardens it. Decoupling lets the refactor land cleanly with the threat model documented, while the hardening lands in a focused PR with its own test coverage.

## References

- `refactor/seinode-drop-sidecar-tls` (commits 2076a80..c51075b)
- `internal/noderesource/noderesource.go` `buildRBACProxyContainer` — `--insecure-listen-address`
- `internal/sidecartransport/sidecartransport.go` — bearer-token injection

PR #267, opened 2026-05-18
refactor(seinode): drop sidecar TLS management; keep kube-rbac-proxy

bdchatham:
## Summary

- Removes the controller's entire sidecar-TLS-management surface. The sidecar API runs on the private K8s cluster network; TLS termination doesn't earn its keep at this layer.
- kube-rbac-proxy stays in front of the sidecar (always-on per SeiNode) for its SAR-based authz value, now in `--insecure-listen-address` HTTP mode.
- Net diff: **−1538 lines** across 27 files after fixups.

## What disappears

- `spec.sidecar.tls` field + `SidecarTLSSpec` type + CEL immutability rule
- `Status.SidecarTLS` field + `SidecarTLSStatus` type + `ConditionSidecarTLSSecretReady` + reason constants
- `internal/controller/node/preflight_sidecar_tls.go` (`reconcileSidecarTLSReady`, `validateTLSSecret`, `requiredDNSNames`, `missingSANs`) + its test file
- `emitSidecarTLSSecretEvent`, TLS-gated periodic requeue in the reconciler
- Planner-side `sidecarTLSSecretReady` gate; init plans always emit `ApplyRBACProxyConfig` (no longer conditional)
- `SidecarTLSEnabled`, `SidecarTLSSecretName`, `sidecarTLSVolumes` helpers, `sidecarTLSVolumeName`/`sidecarTLSMountPath` constants
- `internal/sidecartransport/tls.go` (insecureTLSClientConfig)
- LLD `docs/design-seinode-sidecar-tls-toggle-lld.md`

## What stays

- `kube-rbac-proxy` native sidecar container — now on every SeiNode pod (including bootstrap pods)
- `ApplyRBACProxyConfig` task generates the SAR-policy ConfigMap (group=sei.io, resource=seinodetasks, namespace+name scoped)
- `sidecartransport` bearer-token injection — kube-rbac-proxy still needs the controller SA token for TokenReview
- All other validator-keyring/signing-key/node-key surface unchanged

## What changed in the pod model

- Proxy listens on `:8443` plaintext (`--insecure-listen-address`); SAR authz unchanged
- Sidecar binds loopback unconditionally (`SEI_SIDECAR_AUTHN_MODE=trusted-header`)
- Sidecar liveness/readiness probes are nil — proxy readiness gates pod readiness via upstream forwarding
- Bootstrap pod also gets the proxy (extends the uniform "every sidecar fronted by SAR authz" invariant)
- `SidecarURLForNode` always returns `http://...:8443`
- `cmd/main.go newSidecarClient` always attaches the bearer-token transport (no more spec-derived branching)

## Cross-review

Trio (platform / kubernetes / security) ran against the implementation commit. Two commits address findings:

- `2076a80` — initial refactor (−1576 lines)
- `c51075b` — round-1 fixups (+38/−16)

**Blockers addressed:**
- **K8s**: bootstrap-pod sidecar was reachable on `:7777` while `SidecarURLForNode` had moved to `:8443` → Phase-2 sidecar tasks would hang for snapshot-bootstrap nodes. Fixed by extending the proxy to the bootstrap pod uniformly + moving `ApplyRBACProxyConfig` to Phase 0 so the ConfigMap exists when the bootstrap Job pod schedules.
- **Platform**: `platform.Validate()` didn't require `SEI_KUBE_RBAC_PROXY_IMAGE`. Misconfigured deployments would fail at first reconcile instead of manager startup. Added to the required env map.

**Non-blockers applied:**
- Dropped dead `desired == nil` branch in `apply_rbac_proxy_config.go` (`GenerateRBACProxyConfigMap` no longer returns nil)
- `sidecartransport`: clone `http.DefaultTransport` for process-isolation hygiene
- Cleaned up bootstrap-pod test assertions for the new init-container set

**Deferred (tracked):**
- #265 — NetworkPolicy restricting `:8443` ingress to the controller SA (highest-ROI defense-in-depth)
- #266 — Audit `/v0/metrics` content; now reachable unauthenticated cluster-wide via `--ignore-paths`

## Operator upgrade note

Per the ~10-node internal-only fleet posture: any SeiNode with `spec.sidecar.tls` set today should have that field removed before the controller image bumps. Apiserver will reject writes that include the removed field once the new CRD applies.

```bash
kubectl get seinodes -A -o json | jq -r '.items[] | select(.spec.sidecar.tls != null) | "\(.metadata.namespace) \(.metadata.name)"' \
  | xargs -L1 sh -c 'kubectl -n $0 patch seinode $1 --type=json -p='\''[{"op":"remove","path":"/spec/sidecar/tls"}]'\'''
```

(Per project memory, no production SeiNode has TLS in spec yet, so this is most likely a no-op.)

## Test plan

- [x] `make test` — full suite green
- [x] `golangci-lint run --new-from-rev=origin/main ./...` — 0 issues
- [x] `make manifests generate` — CRD diff is a clean removal of the TLS fields
- [x] Unit: always-on proxy invariants (`sidecar_proxy_test.go` in noderesource + planner)
- [x] Unit: plan composition includes `ApplyRBACProxyConfig` before `ApplyStatefulSet` in all four call sites
- [x] Unit: pod composition (3 init containers, 8 service ports, probes on the proxy)
- [x] End-to-end on harbor: create a SeiNode, verify pod boots cleanly with the proxy in HTTP-only mode and the controller's MarkReady succeeds via bearer-token authn


internal/noderesource/noderesource.go
```diff
@@ -946,122 +932,47 @@ func sidecarSecurityContext() *corev1.SecurityContext {
...
-		},
-		InitialDelaySeconds: 5,
-		PeriodSeconds:       10,
-		FailureThreshold:    3,
-	}
+// Sidecar binds loopback; probes live on the kube-rbac-proxy container.
+// Pod readiness gates on all containers, so a wedged sidecar fails the
+// proxy's readiness probe (upstream returns 5xx) and the pod stays NotReady.
+func sidecarLivenessProbe(_ *seiv1alpha1.SeiNode, _ int32) *corev1.Probe {
+	return nil
```
bdchatham: Remove this?

internal/noderesource/noderesource.go
```diff
@@ -946,122 +932,47 @@ func sidecarSecurityContext() *corev1.SecurityContext {
...
-		ProbeHandler: corev1.ProbeHandler{
-			HTTPGet: &corev1.HTTPGetAction{Path: pathHealthz, Port: intstr.FromInt32(sidecarPort)},
-		},
-		InitialDelaySeconds: 2,
-		PeriodSeconds:       5,
-		FailureThreshold:    6,
-	}
+func sidecarReadinessProbe(_ *seiv1alpha1.SeiNode, _ int32) *corev1.Probe {
+	return nil
 }
```
bdchatham: Remove this?

internal/noderesource/noderesource.go
bdchatham: Done — inlined the nil at the call site and dropped both probe builders in 839b4b4.

internal/noderesource/noderesource.go
bdchatham: Same — 839b4b4.

PR #270, opened 2026-05-18
feat(api): plumb spec.genesis.overrides to sidecar assemble-genesis task

bdchatham:
## Summary

Wires user-supplied genesis parameter overrides from `SeiNodeDeployment.spec.genesis.overrides` into the sidecar's `assemble-and-upload-genesis` task payload, so a fresh test cluster can boot with custom cosmos params (e.g. `staking.params.unbonding_time = 600s`) instead of the cosmos defaults.

## Motivation

SIP-3 testing requires a chain configured for fast-iteration unbonding (10 minutes vs 21 days). The CRD field `spec.genesis.overrides` existed today but was dead-end:

- Its value type was `map[string]string`, which can't carry numbers, durations-as-strings, or nested cosmos param objects without forcing operators to handcraft escaped JSON.
- The controller copied it into a per-node `GenesisCeremonyNodeConfig.GenesisParams` JSON string that **no consumer reads** — the sidecar's `generate-gentx` handler ignores the field, and the assemble-genesis task wasn't given the overrides at all.

The companion seictl PR (see "Linked work" below) extends the sidecar's `AssembleGenesisRequest` with `Overrides map[string]json.RawMessage`; this PR sends the matching field from the controller side.

## API changes

- `SeiNodeDeployment.spec.genesis.overrides` value type reshaped from `map[string]string` to `map[string]apiextensionsv1.JSON`. Keys remain dotted snake_case paths (cosmos-encoded), values are now arbitrary JSON.
- `spec.genesis` is now CRD-immutable after creation (CEL `self == oldSelf`). Genesis-ceremony semantics make in-place edits meaningless; better to fail fast at the API server.
- `GenesisCeremonyNodeConfig.GenesisParams` (per-node JSON string) dropped along with the controller-side `marshalOverrides` helper and the planner's `GenesisParams: gc.GenesisParams` propagation — all three were the dead path.

### Compatibility note

This is a CRD field-type reshape (map value `string` -> JSON), normally a breaking change. Safe in this case because **no live consumer reads the field today** — the sidecar drops it, the controller stamped it onto a per-node field nothing reads. Behavior of existing `SeiNodeDeployment` objects that omit `overrides` is unchanged.

## Wire path

```
spec.genesis.overrides (CRD)
    -> planner ForGroup() (internal/planner/group.go)
        -> task.AssembleAndUploadGenesisTask{Overrides: ...} (internal/task/assemble_genesis.go)
            -> TaskRequest.Params["overrides"] = {...}
                -> seictl sidecar consumes from AssembleGenesisRequest.Overrides
```

`task.AssembleAndUploadGenesisTask` is a thin wrapper that embeds the upstream `sidecar.AssembleAndUploadGenesisTask` and adds `Overrides`. Mirrors the existing `configApplyTask` pattern. Folds back into the upstream type once seictl releases — there's no schema obligation to keep the wrapper after that.

## Test plan

- [x] `make test` passes (added planner-level overrides-propagation + omits-when-unset assertions in `internal/planner/group_test.go`)
- [x] `make manifests generate` regenerates `manifests/crds/sei.io_seinodedeployments.yaml` with the new CEL immutability rule and `x-kubernetes-preserve-unknown-fields: true` on the overrides map values
- [x] Existing assemble-genesis tests still green (`TestBuildGroupAssemblyPlan`, `TestBuildGroupAssemblyPlan_DefaultS3`, `TestBuildPlan_PropagatesAccounts`, etc.)
- [ ] e2e smoke against the SIP-3 test cluster once the companion seictl release ships

## Linked work

- Companion seictl PR adding `Overrides map[string]json.RawMessage` to `AssembleGenesisRequest`: sei-protocol/seictl#181

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

Companion seictl PR: sei-protocol/seictl#181


bdchatham:
Cleanup commit pushed:

- `go.mod` bumped: `sei-protocol/seictl v0.0.47` → `v0.0.50` (seictl#181 + 6 other commits since v0.0.49 — see sei-protocol/seictl@b37c69a8)
- Deleted `internal/task/assemble_genesis.go` (the temporary wrapper)
- Registered `sidecar.AssembleAndUploadGenesisTask` directly in the task registry
- Added `toRawMessages` helper in `internal/planner/group.go` to convert CRD `apiextensionsv1.JSON` values to upstream's `encoding/json.RawMessage` wire type
- Tests updated to read the upstream type

The `go.sum` delta is large because the seictl bump pulled in transitive deps (ProjectZKM/Ziren go-runtime, go-eth-kzg, c-kzg-4844). Not introduced by this PR's logic — go mod transitively resolved them.

Full test suite green; `go vet` clean. Pre-existing golangci-lint findings on `nodes.go` / `networking.go` are unchanged on main.

Issue #360, opened 2026-05-28
Implement publishable P2P (per-pod L4 NLB + ExternalAddress)


bdchatham:
## Problem
Sei pods today are reachable for HTTP (RPC/EVM/REST) via the existing Istio L7 path → shared NLB → private pod, but they are **not publicly dialable as P2P peers** over TCP/26656. This blocks validators (in a deterministic set without sentries) and advertised full nodes from being addressable by ecosystem peers, and prevents internal SNDs from peering each other across clusters.

The design (`platform/docs/designs/sei-publishable-p2p-nlb.md`, merged in `sei-protocol/platform#714`) closes this gap with an opt-in `networking.tcp: {}` sub-struct: when set, the SeiNode reconciler stamps a per-pod `Service type=LoadBalancer` (AWS NLB, `target-type: ip`, scheme=internet-facing, cross-zone=true), reads `Service.status.loadBalancer.ingress[0].hostname`, and writes `SeiNode.Status.ExternalAddress = "<hostname>:26656"`. The planner already reads ExternalAddress into `p2p.external_address` at `planner.go:697`; this issue makes that pipeline reachable.

## Impact
Unblocks:
- **Validator P2P publishability** — deterministic set peers each other over stable NLB DNS endpoints; combined with `unconditional_peer_ids` cross-config (ops, not this issue) gives stable cross-cluster validator mesh.
- **Cross-SND peer discovery** — new `PeerSource.seinodeDeployment` variant resolves sibling SND children's `Status.ExternalAddress` into `persistent_peers`, replacing the EC2-tag-only discovery for K8s-resident publishers.
- **Ecosystem-parity for advertised full nodes** — partner peers can list us as a persistent peer.

Harbor stays excluded until the Cilium overlay is replaced (CGNAT pod IPs aren't VPC-routable → `target-type: ip` fails); controller fails closed via a VPC-CIDR routability check.

## Proposed approach
Six independently-shippable PRs. Sequence dependencies: PR-1 → PR-2 → PR-3 (the chain that makes the field gateable and propagated). PR-4 lands in parallel with PR-3 (independent code path; needs PR-1's types only). PR-5 lands any time before PR-6. PR-6 waits on PR-3 + PR-4 + PR-5.

### PR-1 — Types layer (contract-only, no behavior change)
[...]

PR #361, opened 2026-05-28
feat(api): networking http/tcp sub-structs + seinodeDeployment peer variant (#360 PR-1)

bdchatham:
## Summary

PR-1 of 6 from the publishable-P2P execution plan in #360. Contract-only types layer — no behavior change. None of the new fields are consumed by any controller code yet; subsequent PRs (#360 PR-2 through PR-6) wire them up.

### `NetworkingConfig`
- Adds `HTTP *HTTPConfig` and `TCP *TCPConfig` sub-structs as presence signals for L7 (existing Gateway HTTPRoutes) and L4 (per-pod NLB for P2P/26656) exposure.
- Adds `HTTPEnabled()` accessor with backcompat semantics: legacy `networking: {}` (neither sub-struct set) is treated as `networking: { http: {} }` so existing manifests keep their implicit HTTP exposure. Unit-tested.

### `SeiNodeSpec`
- New `Networking *NetworkingConfig` field. SND template propagation into child SeiNodes lands in PR-2 — left out here.
- Flips `Status.ExternalAddress` doc-comment from "SeiNodeDeployment controller" to "SeiNode controller" and pins the bare `host:port` format (no `nodeId@` prefix, no scheme).

### `PeerSource` union
- New `SeiNodeDeploymentPeerSource` variant for discovering sibling SNDs by label, emitting each child's `Status.ExternalAddress` as a peer entry. XValidation CEL sum bumped to 4.
- Sidecar's `CollectAndSetPeers` task still prepends `nodeID@` at task-build time — same path the `Label` variant feeds, no new sidecar capability.

CRDs + DeepCopy regenerated via `make manifests generate`.

## Backward compatibility

- `networking: nil` → unchanged (no networking).
- `networking: {}` → unchanged behavior via `HTTPEnabled() = true` backcompat; gated by `Networking != nil` everywhere current code looks at it.
- `networking: { http: {} }` → equivalent to `networking: {}` under `HTTPEnabled()`.
- `networking: { tcp: {} }` → field accepted by validation but inert until PR-3 (no consumer yet). Today this would still trigger HTTPRoute stamping at `nodedeployment/networking.go:62` because the gate is nil-check, not `HTTPEnabled()` — PR-3 flips that gate.

## Test plan

- [x] `go test ./api/... ./internal/...` — all green
- [x] `make manifests generate` — CRDs + DeepCopy in sync
- [x] New unit test `TestNetworkingConfig_HTTPEnabled` covers: nil, legacy empty, explicit HTTP, TCP-only, both-set
- [x] `golangci-lint run ./api/v1alpha1/...` — 0 issues
- [x] CI on this PR

## Refs

- Tracking issue: #360
- Design: `sei-protocol/platform` `docs/designs/sei-publishable-p2p-nlb.md`


api/v1alpha1/common_types.go
```diff
@@ -23,6 +23,15 @@ type PeerSource struct {
...
 	Label *LabelPeerSource `json:"label,omitempty"`
+
+	// SeiNodeDeployment discovers peers by selecting sibling
+	// SeiNodeDeployments via labels, then resolves each child
+	// SeiNode's Status.ExternalAddress into a routable host:port.
+	// Use this to peer with publishable nodes that advertise a
+	// per-pod L4 NLB (Spec.Networking.TCP); peers without
+	// ExternalAddress are skipped.
+	// +optional
+	SeiNodeDeployment *SeiNodeDeploymentPeerSource `json:"seinodeDeployment,omitempty"`
```
bdchatham: Please don't add this. You could just use a label peer source that grabs peers by seinodedeployment name label. I'd cut scope.

api/v1alpha1/seinode_types.go
```diff
@@ -36,6 +36,16 @@ type SeiNodeSpec struct {
...
 
+	// Networking opts this node into public networking. HTTP (L7 via
+	// platform Gateway) and TCP (per-pod L4 NLB for P2P/26656) are
+	// independently selectable sub-structs.
+	//
+	// SeiNodes that are children of a SeiNodeDeployment inherit this
+	// from `spec.template.spec.networking` at child-creation time.
+	// Standalone SeiNodes set it directly.
+	// +optional
+	Networking *NetworkingConfig `json:"networking,omitempty"`
```
bdchatham: This seems out of scope.

api/v1alpha1/common_types.go
bdchatham: Agreed — dropped in aedfcd7. PR-3 will extend LabelPeerSource's resolver to prefer Status.ExternalAddress over the headless Service DNS when set, using the existing controller-stamped `sei.io/nodedeployment=<name>` label. No new API surface, no CEL bump.

api/v1alpha1/seinode_types.go
bdchatham: Agreed — dropped in aedfcd7. PR-3's SeiNode reconciler will read the parent SND via OwnerReference for the `Networking.TCP` gate. Single source of truth on the SND, no propagation step needed. Standalone SeiNodes lose publishability in v1; add the field later if a real need arises.

PR #362, opened 2026-05-28
feat(controller/node): publishable P2P — per-pod NLB Service + ExternalAddress gate (#360 PR-2)

bdchatham:
## Summary

PR-2 of the publishable-P2P series (#360). PR-1 (#361) landed the contract; this PR wires the controller path that actually provisions the per-pod NLB Service and writes `Status.ExternalAddress`.

### What this PR delivers

**`internal/controller/node/external_address.go` (NEW)** — `reconcileExternalAddress` runs before `reconcileStatefulSet`:
- **Parent SND lookup** via `metav1.GetControllerOf` (single source of truth on `SND.Spec.Networking.TCP`, no field on `SeiNodeSpec`).
- **Opt-out path** — no TCP → delete Service, clear `Status.ExternalAddress`, `PublishableReady=False/PublishableDisabled`, gate=open.
- **Fail-closed when `SEI_VPC_CIDR` unset** — the safety net for harbor's Cilium CGNAT where `target-type: ip` can't route.
- **Pod-IP routability check** — pod outside CIDR → fail closed; pod absent (first reconcile) → stamp anyway, recheck next round.
- **Per-pod LB Service** server-side-applied with the AWS LBC contract (`type=nlb`, `scheme=internet-facing`, `nlb-target-type=ip`, `cross-zone-load-balancing-enabled=true`), TCP/26656, `externalTrafficPolicy: Local`, selector via `statefulset.kubernetes.io/pod-name=<seinode>-0`.
- **ExternalAddress writer** — `Service.Status.LoadBalancer.Ingress[0].Hostname` → `Status.ExternalAddress = "<host>:26656"`. **Empty-host guard:** don't clear on transient absence; only opt-out clears.

**`internal/controller/node/controller.go`** — gate call before `reconcileStatefulSet`; `Watches(&SeiNodeDeployment{})` with UID-filtered mapper so children re-reconcile when the parent toggles `Networking.TCP`; RBAC for `seinodedeployments` get/list/watch.

**`api/v1alpha1/seinode_types.go`** — `ConditionPublishableReady` + 5 reason constants.

**`internal/controller/nodedeployment/{networking,status}.go`** — three call sites migrated from `Networking == nil` to `!Networking.HTTPEnabled()`. Without this, manifests setting only `networking.tcp: {}` would spuriously stamp HTTPRoutes + ClusterIP from the SND reconciler.

### Backward compatibility

- `networking: nil` — unchanged.
- `networking: {}` — unchanged (HTTPEnabled=true via backcompat).
- `networking: { http: {} }` — unchanged (HTTPEnabled=true).
- `networking: { tcp: {} }` — **now correctly does NOT trigger HTTPRoutes** (was a latent bug after PR-1 if anyone set this).
- `networking: { http: {}, tcp: {} }` — both layers active.

### Test plan

- [x] 12 unit tests in `external_address_test.go` — standalone, opt-out, opt-out cleanup, VPC unset (fail-closed), pod outside CIDR (fail-closed), Service stamping + gate held closed, hostname populated, empty-host guard, mapper.
- [x] Envtest `TestPublishableP2P_HoldsGateUntilHostnamePopulates` — full bootstrap-gate happy path + opt-out cleanup transition.
- [x] `go test ./api/v1alpha1/... ./internal/...` — all green.
- [x] `make test-integration` — envtest suite green (59s).
- [x] `golangci-lint run ./internal/controller/node/... ./internal/controller/nodedeployment/... ./api/v1alpha1/...` — no new issues.
- [x] `make manifests generate` — no diff (no new CRD spec fields; new RBAC marker collapsed into existing role).
- [ ] CI on this PR.

### Out of scope (separate follow-ups)

- **`LabelPeerSource` resolver enhancement** — make it prefer `Status.ExternalAddress` over headless DNS when set (per the scope cut on #361). Tracked as part of #360.
- **Terraform SG opening** for `tcp/26656 from 0.0.0.0/0` — platform PR.
- **`SEI_VPC_CIDR` env var on `manager-patch.yaml`** — platform PR. Until that lands, the feature is inert (condition surfaces `VPCCIDRNotConfigured`, gate held closed). Safe to merge this PR independently.

### Refs

- Tracking issue: #360
- Design: `sei-protocol/platform` `docs/designs/sei-publishable-p2p-nlb.md`
- Prior PR (types contract): #361

### DIFF
--- a/api/v1alpha1/seinode_types.go
+++ b/api/v1alpha1/seinode_types.go
@@ -277,6 +277,31 @@ const (
 
 	// ConditionSeiNodePaused mirrors spec.paused: True when paused.
 	ConditionSeiNodePaused = "Paused"
+
+	// ConditionPublishableReady reports the per-pod L4 NLB state.
+	// True when Status.ExternalAddress is populated; False with reason
+	// carries the off-or-not-yet state.
+	ConditionPublishableReady = "PublishableReady"
+)
+
+// Reasons for the PublishableReady condition.
+const (
+	// ReasonPublishableDisabled — parent SND has no Networking.TCP.
+	ReasonPublishableDisabled = "PublishableDisabled"
+
+	// ReasonVPCCIDRNotConfigured — fail-closed: SEI_VPC_CIDR is unset.
+	ReasonVPCCIDRNotConfigured = "VPCCIDRNotConfigured"
+
+	// ReasonPodOutsideVPCCIDR — fail-closed: pod IP outside the CIDR
+	// (harbor's Cilium CGNAT). NLB target-type=ip needs routable IPs.
+	ReasonPodOutsideVPCCIDR = "PodOutsideVPCCIDR"
+
+	// ReasonServiceProvisioning — Service exists, NLB hostname not yet
+	// populated (~2 min on AWS).
+	ReasonServiceProvisioning = "ServiceProvisioning"
+
+	// ReasonExternalAddressReady — Status.ExternalAddress is set.
+	ReasonExternalAddressReady = "ExternalAddressReady"
 )
 
 // Reasons for the ImportPVCReady condition.
--- a/internal/controller/node/controller.go
+++ b/internal/controller/node/controller.go
@@ -20,8 +20,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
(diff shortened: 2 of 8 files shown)


bdchatham:
Closing in favor of an architectural reframing surfaced during peer review.

The current PR puts publishable-P2P networking ownership in the SeiNode reconciler, which is the wrong layer — networking is consistently a SeiNodeDeployment concern in this codebase (HTTPRoutes and the internal ClusterIP are both SND-owned). The per-pod LB Service should live with them. The current PR also took a raw NLB hostname rather than an external-dns-fronted deterministic vanity hostname, which forced the bootstrap Service-ready gate and the per-reconcile validation gates. With a deterministic hostname known at SND-create time, the gate dissolves.

Replacement PR will:
- Move per-child Service stamping to the SND reconciler
- Use external-dns + a deterministic ``<seinode>-p2p.platform.sei.io`` hostname so ``Status.ExternalAddress`` is known up front
- Drop ``runExternalAddressGate`` and the parent-SND lookup
- Lift ``SEI_VPC_CIDR`` validation from per-reconcile to startup (or admission)

The call-site migration from ``Networking == nil`` to ``!HTTPEnabled()`` will land as its own small PR since it is cleanly correct standalone.

Tracking on #360.

PR #363, opened 2026-05-28
fix(load-test/rpc): enable tx_index.indexer=kv

bdchatham:
Chaos suite seiload logs showed a steady stream of `❌ timeout waiting for receipt for tx 0x...` even though the chain was committing transactions. Root cause: the load-test RPC template doesn't set `tx_index.indexer`, so seid doesn't index txs by hash and `eth_getTransactionReceipt` can't find them. release-test/rpc.yaml.tmpl has had this setting since inception.

### DIFF
--- a/scenarios/load-test/rpc.yaml.tmpl
+++ b/scenarios/load-test/rpc.yaml.tmpl
@@ -14,6 +14,7 @@ spec:
             selector:
               sei.io/chain: "{{ .CHAIN_ID }}"
       overrides:
+        tx_index.indexer: kv
         storage.state_commit.write_mode: memiavl_only
         storage.state_store.write_mode: memiavl_only
   updateStrategy:


bdchatham:
Closing — verified against sei-chain source (cited file:line refs from platform-engineer deep dive). The `tx_index.indexer: kv` override is **a no-op** for `eth_getTransactionReceipt`:

- EVM receipts live in sei-db's separate `receipt.db` (`x/evm/keeper/receipt.go`), not CometBFT's tx_index
- The receipt store is **unconditionally** initialized for every node (`app/app.go:685-695`)
- `sei-config DefaultForMode(ModeFull)` already sets `tx_index.indexer = kv` (`defaults.go:288`) — full nodes get this without override
- The release-test template's `tx_index.indexer: kv` line is also redundant

The actual cause of seiload receipt timeouts is EVM RPC **worker pool saturation** under 2-4k tps load (`evm.worker_pool_size = min(64, NumCPU*2)` defaults to 8 on a 4-vCPU pod). Filing a different PR with the right tuning.

Issue #368, opened 2026-05-29
Harmonize SeiNode peer discovery between controller and sidecar


bdchatham:
# Problem

Peer discovery + config writing for SeiNodes is split across two controllers today, and the split is not principled:

- **`internal/controller/node/peers.go`** (controller-side) — `reconcilePeers` resolves `Spec.Peers` (EC2 tags, K8s labels, static) into a list of `host:port` entries written to `Status.ResolvedPeers`. Runs every reconcile.
- **`CollectAndSetPeers`** (sidecar-side) — runs as a planned task. Reads `Status.ResolvedPeers` from its own SeiNode, queries each peer's `:26657/status` to discover the peer's `node_id`, then writes the full `<nodeID>@<host>:<port>` list into `persistent_peers` in `config.toml` and reloads seid.

The split is "controller discovers K8s/AWS membership; sidecar enriches with on-chain identity." That's a defensible separation in isolation, but in practice it means:

1. Two reconcile cadences for one logical concern — controller-side runs every reconcile (~30s steady-state); sidecar-side runs only when the planner schedules `CollectAndSetPeers`.
2. Two error surfaces — controller errors land on SND/SeiNode status; sidecar errors land on task status. An operator chasing "why isn't this node peering?" has to check both.
3. Membership drift between the two — `Status.ResolvedPeers` can be stale relative to what the sidecar actually wrote into `persistent_peers`.
4. The `seinodeDeployment` PeerSource extension we landed in PR-1 (#361, then dropped during PR-3 cuts in favor of using the existing `LabelPeerSource` with the `sei.io/nodedeployment` label) is a hint that the controller-side resolver is already doing more than just "K8s membership" — it's resolving cross-SND peer relationships that fold into the same path.

# Impact

This shows up immediately when we go to roll out publishable P2P (#360) and start exercising the cross-cluster peer-discovery story. Validators-peering-validators across a deterministic set means the controller's resolver becomes the load-bearing membership service, and the split-with-sidecar-enrichment adds enough latency + opacity that it becomes a debugging hot spot during P2P bring-up.

# Proposed approach (to refine)

Three shapes worth considering:

1. **Controller owns everything** — controller resolves peers AND queries `node_id` (proxy through sidecar HTTP if needed), writes the final `persistent_peers` directly. Sidecar's `CollectAndSetPeers` becomes config-apply only.
[...]

PR #365, opened 2026-05-28
feat(p2p-endpoint): SND-owned per-pod NLB + Spec.ExternalAddress (#360)

bdchatham:
## Summary

Reframe of the publishable-P2P controller path (#360). Supersedes #362.

## Design

**Networking is an SND-layer concern.** The SND reconciler owns the per-pod LoadBalancer Service (one per child SeiNode), paralleling the existing HTTPRoute path. The SeiNode reconciler is fully networking-agnostic — it reads `Spec.ExternalAddress` and proceeds.

**`Spec.ExternalAddress` is plain `string` on `SeiNodeSpec`.** The SND computes the deterministic vanity hostname per child and injects the address at child Create time via the `ensureSeiNode` diff path. Standalone SeiNodes (no SND parent) can set the field directly; the controller respects user-set values on standalone resources.

**Deterministic vanity hostname via external-dns.** Format: `<seinode-name>-p2p.<chainID>.<P2PEndpointDomain>`. The SND stamps `external-dns.alpha.kubernetes.io/hostname` on each per-pod Service; external-dns creates the CNAME to the AWS-allocated NLB hostname. ChainID has a DNS-1123 pattern constraint on the CRD; that's the gate.

**`SEI_P2P_ENDPOINT_DOMAIN` env, parsed once at startup.** When unset, the P2P endpoint path is silently inert — TCP-enabled SNDs see no Service stamped and no `Spec.ExternalAddress` set. `ConditionNetworkingReady=False/NetworkingDisabled` reflects "no tier active" so operators have a clear signal during cluster bring-up. CNI mismatches surface via AWS LBC target-registration failure events, not a controller-side gate.

**Lifecycle.** Opt-in stamps `Spec.ExternalAddress` + creates the per-pod Service. Opt-out clears the field and deletes the Service. Scale-down deletes the per-ordinal Service before the child SeiNode. `DeletionPolicyRetain` orphans the per-pod Services alongside the HTTPRoute / ClusterIP. Runtime opt toggles require a manual pod delete to re-render `config.toml` — v1 ops contract.

## Files

- `api/v1alpha1/seinode_types.go` — drop `Status.ExternalAddress`; add `Spec.ExternalAddress string`; ChainID DNS-1123 pattern.
- `api/v1alpha1/seinodedeployment_types.go` — `NetworkingStatus.P2PEndpoints []P2PEndpoint` (ordinal-keyed list); Genesis.ChainID DNS-1123 pattern.
[...]

PR #369, opened 2026-05-29
feat(controller/node): resolve label peers to NLB addresses + controller-side node_id (#368)

bdchatham:
## Summary

First slice of the peer-discovery harmonization (#368). Moves node_id resolution for the `LabelPeerSource` flow from the sidecar's `:26657/status` query to the controller's existing per-peer sidecar gRPC. `Status.ResolvedPeers` shifts to fully-composed `<node_id>@<host>:<port>` strings.

## Why

The sidecar's `:26657/status` lookup path is in-cluster-only — NLBs only forward 26656, so it can't reach a publishable peer over the public endpoint. After PR #365, a peer's `Spec.ExternalAddress` is the publishable vanity hostname; if the controller writes that into `Status.ResolvedPeers` and the sidecar tries `:26657/status`, the lookup fails permanently.

The controller already has in-cluster access to each peer's sidecar gRPC (port 7777) — the same path the genesis `CollectAndSetPeers` task uses to read `node_id` from `node_key.json`. Reusing that path for the Label flow keeps node_id resolution authoritative (it's from the keyfile, not from `:26657`) and decouples discovery from publishable-vs-private addressing.

## Behavior

| Peer's `Spec.ExternalAddress` | Resolved entry |
|---|---|
| set (`<vanity>:26656`) | `<node_id>@<vanity>:26656` |
| unset | `<node_id>@<peer>-0.<peer>.<ns>.svc.cluster.local:26656` |

`node_id` is fetched via the existing `SidecarClient.GetNodeID` (in-cluster gRPC to port 7777). On per-peer failure the whole resolve errors and controller-runtime retries — same atomic semantic as the genesis `CollectAndSetPeers` task. Seid only reads `persistent_peers` at boot, so transient peer-sidecar failures during reconcile don't disturb running pods.

## Files

- `internal/controller/node/peers.go` — `resolveLabelPeers` composes the new wire format; new `peerAddress` helper picks `Spec.ExternalAddress` or falls back to headless DNS.
- `internal/planner/planner.go` — Label sources map to `PeerSourceStatic` instead of `PeerSourceDNSEndpoints`. The sidecar's `DNSEndpointsSource` is untouched — still serves `EC2Tags` peers.
- `internal/task/task.go` — widens `task.SidecarClient` interface with `GetNodeID(ctx) (string, error)`. The full `*sidecar.SidecarClient` already implements it; mocks gain one-line stubs.
- `api/v1alpha1/seinode_types.go` — `ResolvedPeers` godoc updated for the new composed wire format.

## Wire-format change (one-way door)

`Status.ResolvedPeers` was previously a `[]string` of bare hosts (`<peer>-0.<peer>.<ns>.svc.cluster.local`). It's now a `[]string` of fully-composed `<node_id>@<host>:<port>` entries. The only consumer is the planner, which is updated in this same PR. No external consumer is affected.

## Backward compatibility

- Existing SNDs with `LabelPeerSource` keep working — their `ResolvedPeers` will repopulate with the new format on next reconcile, and the planner routes them through the same `Static` sidecar source path that genesis peers already use.
- `Spec.Peers` user-facing shape is unchanged.
- `EC2Tags` and `Static` PeerSources are unchanged.
- The sidecar's `DNSEndpointsSource` (the `:26657/status` path) stays in `seictl` — still serves `EC2Tags`. A follow-up could retire it once `EC2Tags` also moves to controller-side; out of scope here.

## Test plan

- [x] Unit: new `TestReconcilePeers_PrefersExternalAddress`. Existing peers tests updated for the new format.
- [x] `go test ./api/v1alpha1/... ./internal/...` — green.
- [x] `make test-integration` — envtest green (~64s).
- [x] `golangci-lint run --new-from-rev=main ./...` — 0 issues.
- [x] `make manifests generate` — clean.
- [ ] CI on this PR.

## Refs

- Tracking issue: #368
- Prior PR (controller publishable path): #365
- Platform follow-ups: `sei-protocol/platform#747` (env wiring), `sei-protocol/platform#748` (SG opening)

### DIFF
--- a/api/v1alpha1/seinode_types.go
+++ b/api/v1alpha1/seinode_types.go
@@ -347,10 +347,8 @@ type SeiNodeStatus struct {
 	// +optional
 	Plan *TaskPlan `json:"plan,omitempty"`
 
-	// ResolvedPeers is the current set of `<node_id>@<host>:<port>` peer
-	// strings discovered from label-based peer sources, ready to write
-	// into CometBFT's persistent_peers verbatim. Host is the peer's
-	// Spec.ExternalAddress when set, otherwise the headless Service DNS.
+	// ResolvedPeers carries `<node_id>@<host>:<port>` entries resolved
+	// from label-based peer sources, ready for CometBFT's persistent_peers.
 	// +optional
 	ResolvedPeers []string `json:"resolvedPeers,omitempty"`
 
--- a/api/v1alpha1/seinode_types.go
+++ b/api/v1alpha1/seinode_types.go
@@ -347,9 +347,10 @@ type SeiNodeStatus struct {
 	// +optional
 	Plan *TaskPlan `json:"plan,omitempty"`
 
-	// ResolvedPeers is the current set of peer DNS hostnames discovered
-	// from label-based peer sources. Reconciled continuously so that
-	// future peer-update plans can detect drift.
+	// ResolvedPeers is the current set of `<node_id>@<host>:<port>` peer
+	// strings discovered from label-based peer sources, ready to write
+	// into CometBFT's persistent_peers verbatim. Host is the peer's
+	// Spec.ExternalAddress when set, otherwise the headless Service DNS.
 	// +optional
 	ResolvedPeers []string `json:"resolvedPeers,omitempty"`
 
--- a/config/crd/sei.io_seinodes.yaml
+++ b/config/crd/sei.io_seinodes.yaml
@@ -957,9 +957,10 @@ spec:
                 type: object
               resolvedPeers:
                 description: |-
-                  ResolvedPeers is the current set of peer DNS hostnames discovered
-                  from label-based peer sources. Reconciled continuously so that
-                  future peer-update plans can detect drift.
+                  ResolvedPeers is the current set of `<node_id>@<host>:<port>` peer
+                  strings discovered from label-based peer sources, ready to write
+                  into CometBFT's persistent_peers verbatim. Host is the peer's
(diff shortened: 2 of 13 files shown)


bdchatham:
Three-way cross-review verdict:
- **kubernetes-specialist:** GO (atomic semantics correct)
- **sei-network-specialist:** GO (CometBFT semantics + node_id correctness verified)
- **platform-engineer:** HOLD — atomic resolve wedges fleet-wide reconciles in steady state when any label-matched peer's sidecar is transient-unavailable (pacific-1 scenario: 6+ SNDs with mutual selectors → single sidecar restart blocks every consumer's reconcile, not just peer updates).

Reasoning from first principles per coral discipline: atomic is right for the genesis path (bounded cohort, one-shot, co-bootstrapping) but wrong for the steady-state label flow (every reconcile forever, arbitrary peer lifecycle). Platform-engineer's blocker is the load-bearing one.

Fix pushed in `0fceb7a`:
- On per-peer BuildSidecarClient or GetNodeID error, preserve that peer's prior entry from `Status.ResolvedPeers` (indexed by host:port). Transients don't drop peers from the list.
- New peer with no prior entry + sidecar failure: skip with structured log line (`peer`, `err`), do not fail the reconcile. Mitigates the k8s-specialist's "silent partial list" debugging concern.
- Two new unit tests: `TestReconcilePeers_PreservesPriorEntryOnTransientFailure`, `TestReconcilePeers_SkipsNewPeerOnSidecarFailure`.

Tests + envtest still green. `golangci-lint --new-from-rev=main` = 0 issues.

PR #374, opened 2026-05-30
feat(planner): per-mode update plans with config-patch

bdchatham:
## Problem

When a SeiNode is in `Running` phase, the controller's update plan rolls the pod but never re-applies `config.toml`. Any controller-owned field that flips after init — `Spec.ExternalAddress` from the publishable-P2P feature being the load-bearing example — stays empty on the pod. Inbound NLB → pod works; outbound advertising (NodeInfo / PEX) doesn't, because seid reads `external-address = ''` from the stale config.

Caught during the arctic-1 canary rollout (platform#764–771). With AWS LBC + harbor `SEI_NLB_TARGET_TYPE=instance` in place, the NLB path is live, but seid still doesn't tell other peers how to reach it.

## Fix

Each mode planner's `BuildPlan` now dispatches internally between init and a Running node. The Running branch emits an update plan with a `TaskConfigPatch` between `ApplyService` and `ReplacePod`:

```
[validator only: ValidateSigningKey → ValidateNodeKey → ValidateOperatorKeyring →]
ApplyStatefulSet → ApplyService → TaskConfigPatch → TaskConfigValidate →
ReplacePod → ObserveImage → MarkReady
```

`TaskConfigPatch` is a generic TOML merge — read the file, merge the controller-named keys, write back. No sei-config involvement. The patch the controller emits today is literally `{"config.toml": {"p2p": {"external-address": Spec.ExternalAddress}}}`. `TaskConfigValidate` after the patch is the parser-level gate (sei-config's typed Diagnostics are deliberately not re-run on the day-2 path — keeps the controller decoupled from sei-config's typed registry).

The validator's progression prepends the three key-validation gates so a missing or malformed secret aborts before any STS mutation (Cursor Bugbot finding).

## Design notes the iteration converged on

- **`NodePlanner` interface unchanged** from main: `Validate`, `BuildPlan`, `Mode`. Reconciler/executor just call `BuildPlan` and get a plan. Earlier iterations lifted `BuildConfigIntent` onto the interface; user feedback was that lifting config-intent out made it a special case. Each mode planner now owns its full update progression inline.
- **Init writes the whole config via `TaskConfigApply` (sei-config mode-defaulted). An existing running resource patches only controller-owned keys via `TaskConfigPatch`.** Never both in one plan. Stated as a doc rule on `NodePlanner` and pinned by `TestPlannerConvention_InitUsesApply_UpdateUsesPatch`.
- **`assembleUpdatePlan` is pure boilerplate** (planID, task ordering, param marshaling, phase fields). Callers own the `NodeUpdateInProgress` condition — they stamp it with the reason/message that reflects their trigger before calling. Keeps the helper generic for future Running-phase triggers without a misleading "image drift detected" message.
- **`pelletier/go-toml/v2` does not preserve comments** on re-encode. First config-patch on a PVC erases operator-added comments. Acknowledged in `fullNodePlanner.buildRunningPlan`.

## Tests

- `TestFullPlanner_ImageDrift_UpdateProgression` — 7-task ordering for full nodes.
- `TestArchivePlanner_ImageDrift_UpdateProgression`, `TestReplayerPlanner_ImageDrift_UpdateProgression` — symmetry coverage.
- `TestValidatorPlanner_ImageDrift_PrependsValidationGates` — 10-task ordering with the three validation gates.
- `TestFullPlanner_ConfigPatchCarriesExternalAddress` — patch payload propagates `Spec.ExternalAddress`.
- `TestFullPlanner_ConfigPatchIsDeterministic` — same spec → same patch.
- `TestPlannerConvention_InitUsesApply_UpdateUsesPatch` — init=Apply, update=Patch, never both.
- `TestConfigPatchTask_JSONWireFormatUsesFilesKey` — pins the `json:\"files\"` tag (drop it and every patch silently no-ops at the sidecar).
- Existing `TestFullPlanner_Sidecar*`, `TestFullPlanner_ImageDriftWinsOverSidecar`, condition-lifecycle tests adapted to the new shape.

Unit + envtest + lint all green.

## Out of scope

Autonomous `Spec.ExternalAddress` drift detection on a Running node. Today the patch propagates on the next trigger of `buildRunningPlan` (image drift; operator can SHA-pin to force one). If NLB hostname rotation or `Spec.Overrides` drift become operationally painful, a follow-up adds an external-address drift check that routes to the same update progression.

## Test plan

- [x] CI green (unit, envtest, lint).
- [x] After merge + ECR build + harbor controller image bump: SHA-pinning arctic-1 syncer image triggers `buildRunningPlan` → `config.toml` on the new pod shows `external-address = \"syncer-0-0-p2p.arctic-1.harbor.platform.sei.io:26656\"` → seid logs show PEX advertising the NLB hostname.

### DIFF
--- a/internal/planner/full.go
+++ b/internal/planner/full.go
@@ -44,6 +44,9 @@ func (p *fullNodePlanner) BuildPlan(node *seiv1alpha1.SeiNode) (*seiv1alpha1.Tas
 }
 
 func (p *fullNodePlanner) BuildConfigIntent(node *seiv1alpha1.SeiNode) (*seiconfig.ConfigIntent, error) {
+	if node.Spec.FullNode == nil {
+		return nil, fmt.Errorf("fullNode sub-spec is nil")
+	}
 	return &seiconfig.ConfigIntent{
 		Mode:      seiconfig.ModeFull,
 		Overrides: mergeOverrides(mergeOverrides(commonOverrides(node), p.controllerOverrides(node)), node.Spec.Overrides),
--- a/internal/planner/full.go
+++ b/internal/planner/full.go
@@ -33,14 +33,21 @@ func (p *fullNodePlanner) BuildPlan(node *seiv1alpha1.SeiNode) (*seiv1alpha1.Tas
 		return buildRunningPlan(node)
 	}
 	fn := node.Spec.FullNode
-	params := &seiconfig.ConfigIntent{
-		Mode:      seiconfig.ModeFull,
-		Overrides: mergeOverrides(mergeOverrides(commonOverrides(node), p.controllerOverrides(node)), node.Spec.Overrides),
+	params, err := p.BuildConfigIntent(node)
+	if err != nil {
+		return nil, err
 	}
 	if NeedsBootstrap(node) {
 		return buildBootstrapPlan(node, node.Spec.Peers, fn.Snapshot, params)
 	}
 	return buildBasePlan(node, node.Spec.Peers, fn.Snapshot, params)
+}
+
+func (p *fullNodePlanner) BuildConfigIntent(node *seiv1alpha1.SeiNode) (*seiconfig.ConfigIntent, error) {
+	return &seiconfig.ConfigIntent{
+		Mode:      seiconfig.ModeFull,
+		Overrides: mergeOverrides(mergeOverrides(commonOverrides(node), p.controllerOverrides(node)), node.Spec.Overrides),
+	}, nil
 }
 
 func (p *fullNodePlanner) controllerOverrides(node *seiv1alpha1.SeiNode) map[string]string {
--- a/internal/planner/full.go
+++ b/internal/planner/full.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	seiconfig "github.com/sei-protocol/sei-config"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	seiv1alpha1 "github.com/sei-protocol/sei-k8s-controller/api/v1alpha1"
 	"github.com/sei-protocol/sei-k8s-controller/internal/task"
@@ -49,6 +50,7 @@ func (p *fullNodePlanner) BuildPlan(node *seiv1alpha1.SeiNode) (*seiv1alpha1.Tas
 // re-encode — the first config-patch erases operator-added comments.
 func (p *fullNodePlanner) buildRunningPlan(node *seiv1alpha1.SeiNode) (*seiv1alpha1.TaskPlan, error) {
 	if imageDrifted(node) {
+		setNodeUpdateCondition(node, metav1.ConditionTrue, "UpdateStarted", imageDriftMessage(node))
 		prog := []string{
 			task.TaskTypeApplyStatefulSet,
 			task.TaskTypeApplyService,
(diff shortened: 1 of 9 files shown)


internal/planner/planner.go
```diff
@@ -735,18 +742,44 @@ func buildMarkReadyPlan(node *seiv1alpha1.SeiNode) (*seiv1alpha1.TaskPlan, error
...
+// the non-incremental applyFull, which regenerates config from mode
+// defaults + overrides; running that on day 2 would wipe persistent-peers,
+// state-sync trust point, and any operator-managed TOML keys that the
+// controller doesn't track in overrides. applyIncremental reads on-disk
+// config and patches only the overrides we own.
 func buildNodeUpdatePlan(node *seiv1alpha1.SeiNode) (*seiv1alpha1.TaskPlan, error) {
 	setNodeUpdateCondition(node, metav1.ConditionTrue, "UpdateStarted",
 		fmt.Sprintf("image drift detected: spec=%s current=%s", node.Spec.Image, node.Status.CurrentImage))
 
+	mode, err := plannerForMode(node)
```
bdchatham: can we refactor mode -> nodePlanner? Mode is very misleading but I understand we want to avoid shadowing.

internal/planner/planner.go
```diff
@@ -735,18 +742,44 @@ func buildMarkReadyPlan(node *seiv1alpha1.SeiNode) (*seiv1alpha1.TaskPlan, error
...
+	if err != nil {
+		return nil, fmt.Errorf("resolving mode planner: %w", err)
+	}
+	intent, err := mode.BuildConfigIntent(node)
```
bdchatham: Having a separate method for this is a smell. We should be able to determine this in planners

internal/planner/full.go
```diff
@@ -33,16 +33,26 @@ func (p *fullNodePlanner) BuildPlan(node *seiv1alpha1.SeiNode) (*seiv1alpha1.Tas
...
+func (p *fullNodePlanner) BuildConfigIntent(node *seiv1alpha1.SeiNode) (*seiconfig.ConfigIntent, error) {
+	if node.Spec.FullNode == nil {
+		return nil, fmt.Errorf("fullNode sub-spec is nil")
+	}
+	return &seiconfig.ConfigIntent{
+		Mode:      seiconfig.ModeFull,
+		Overrides: mergeOverrides(mergeOverrides(commonOverrides(node), p.controllerOverrides(node)), node.Spec.Overrides),
+	}, nil
+}
+
```
bdchatham: We shouldn't need a new method to determine config intent. In my ideal scenario, we determine this in the planner itself. It's okay if we do this in multiple places, repeating code. Doing it this way makes config intent a special case, which it's really just another task/sequence of tasks

internal/planner/planner.go
```diff
@@ -734,28 +736,48 @@ func buildMarkReadyPlan(node *seiv1alpha1.SeiNode) (*seiv1alpha1.TaskPlan, error
...
+// it needs (e.g. validator prepends key-validation gates).
+//
+// patch carries the TOML fragments the controller wants stamped into
+// named files for TaskConfigPatch; pass nil if the planner's progression
+// doesn't include a config-patch step (none today).
 //
 // FailedPhase is deliberately empty: a failure retries on the next reconcile
 // rather than transitioning the node out of Running.
-func buildNodeUpdatePlan(node *seiv1alpha1.SeiNode) (*seiv1alpha1.TaskPlan, error) {
+func assembleDay2Plan(node *seiv1alpha1.SeiNode, prog []string, patch map[string]map[string]any) (*seiv1alpha1.TaskPlan, error) {
```
bdchatham: Let's avoid unconventional naming like "day 2"

PR #384, opened 2026-06-03
fix(statesync): resolve internal-RPC witnesses for label peers

bdchatham:
## Problem
State-sync light-client witnesses were derived **by the sidecar** from `persistent_peers`. For `networking.tcp` peers, `peerAddress` returns `Spec.ExternalAddress` — the external **P2P** NLB hostname, which exposes 26656 only and has no RPC listener. The sidecar's `extractRPCHosts` stripped the port and appended `:26657`, producing a witness on a host that never answers `/status`. seid exited on `no witnesses connected` and crashlooped. This is the regression that took arctic-1 node-19 down during the BYO-key migration (manually patched in prod; this codifies the fix).

## Fix (controller half of the A+B fix)
Resolve a **parallel witness list** from the same label-selected peers, using each peer's in-cluster headless RPC DNS and never the external address:

- New `peerRPCAddress(peer)` → `<peer>-0.<peer>.<ns>.svc.cluster.local:26657` (`seiconfig.PortRPC`). Unlike `peerAddress` it never consults `Spec.ExternalAddress` — RPC is internal-only.
- `reconcilePeers`/`resolveLabelPeers` collect witnesses alongside `ResolvedPeers` and store them in a new `Status.ResolvedRPCWitnesses`.
- `configureStateSyncTask` passes them to `ConfigureStateSyncTask.RpcServers` (requires seictl **v0.0.55**, bumped here).

Witnesses are deterministic from peer identity (no `node_id`, no sidecar call), so every matched peer yields a witness **regardless of sidecar reachability** — strictly more robust than peer resolution. Empty (EC2/static-peer nodes) leaves the sidecar to derive witnesses from `persistent_peers` as before.

This pairs with seictl #197 (merged, v0.0.55): the sidecar uses explicit `RpcServers` verbatim when provided and `/status`-probes each, dropping unreachable ones.

## Changes
- `internal/controller/node/peers.go` — `peerRPCAddress` + witness collection
- `api/v1alpha1/seinode_types.go` — `Status.ResolvedRPCWitnesses` (+ regenerated deepcopy + CRD)
- `internal/planner/planner.go` — pass witnesses into the configure-state-sync task
- `go.mod` — seictl v0.0.50 → v0.0.55

## Tests
- `TestReconcilePeers_PrefersExternalAddress` — extended: external-P2P peer still yields **internal RPC DNS** witness (the regression guard)
- `TestReconcilePeers_WitnessesExcludeSelfAndUseRPCPort`
- `TestConfigureStateSyncTask_PassesResolvedWitnesses` / `_NoWitnessesLeavesEmpty`


internal/planner/planner.go
```diff
@@ -609,7 +609,7 @@ func paramsForTaskType(
 	case TaskDiscoverPeers:
 		return discoverPeersTask(node)
 	case TaskConfigureStateSync:
-		return configureStateSyncTask(snap)
+		return configureStateSyncTask(node, snap)
```
bdchatham: if you pass in the node you could just source the snapshot source from it.

bdchatham: Good call — done in a6612e6. `configureStateSyncTask(node)` now derives `snap := node.Spec.SnapshotSource()` internally instead of taking the redundant arg. Verified equivalent for every node type: full/validator/replay callers already passed exactly `SnapshotSource()`, and archive uses `Spec.Archive` (not in the switch → nil), matching the explicit nil archive.go passed.

PR #386, opened 2026-06-06
feat(nodetask): add DiscoverPeers + RestartPod SeiNodeTask kinds

bdchatham:
## What

Two new composable `SeiNodeTask` kinds to refresh a **running** node's `persistent-peers` without baking peer-drift logic into the core `SeiNode` reconcile:

- **`DiscoverPeers`** (sidecar-backed) — re-resolves the target SeiNode's `spec.peers` and writes `persistent-peers` into `config.toml` on the PVC via the existing sidecar `discover-peers` task. Scalar merge → preserves the rest of the config, so **no `config-apply` needed**. Disk-only: the running seid picks it up on its next restart.
- **`RestartPod`** (controller-side) — restarts the target's pod so seid re-reads `config.toml`.

Sequenced **`DiscoverPeers` → `RestartPod`** applies fresh peers to a live node (the arctic-1 EC2→K8s validator migration use case), composable into Chaos Mesh Workflows.

## Why

Migrated K8s validators need their `persistent-peers` updated after a `spec.peers` change (e.g. adding the proposer cohort). We established that:
- `peers` is **not** in the controller's rollout `templateHash`, so a peers edit doesn't trigger a plan;
- the NodeUpdate (image) plan runs neither `discover-peers` nor `config-apply`, and `mark-ready` is a no-op — so an image bump restarts seid onto the **stale** `config.toml`;
- there is **no working hot-reload** of `persistent-peers` into a live seid (seictl `config-reload`'s seid-signal is a `TODO`).

Rather than enshrine a peer-drift trigger in the core reconcile (a heavier, harder-to-reverse change with fleet-wide blast radius), this delivers the capability as **opt-in, composable tasks**. Promote to an automatic controller trigger later if it proves worth it (tracked separately).

## Design notes

- **`RestartPod` completion is content-addressed by pod UID.** Execute deletes only the captured-UID pod; Status completes when an owned `Ready` pod has a *different* UID. The UID is captured at synthesis and persisted on `status.task` (survives reconcile reconstruction). This is immune to the `creationTimestamp` 1s-resolution race a wall-clock epoch would have. It does **not** reuse `replace-pod` (which keys on StatefulSet revision and no-ops when the spec is unchanged).
- **Bounded by default.** `RestartPod` defaults to a 10m timeout, `DiscoverPeers` 2m (explicit `spec.timeoutSeconds` always wins; sidecar gov/await kinds stay unbounded). A hand-authored `RestartPod` whose pod never goes Ready transitions to `Failed(reason=Timeout)` rather than requeuing forever.
- **Fail-fast on empty peers.** A target with no `spec.peers` — including a `label` source whose `status.resolvedPeers` is empty — fails with `ParamsBuildFailed` rather than submitting a zero-source `discover-peers` that would silently wipe `persistent-peers`.
- **Double-sign safe.** `RestartPod` is the same single-replica + OnDelete + RWO-PVC stop-then-start as the existing `replace-pod`; multi-replica is a terminal guard.
- **Atomicity caveat (documented on the kind):** `DiscoverPeers` succeeds then `RestartPod` fails → fresh config on disk + old seid still running; transient, observable, self-healing on the next successful `RestartPod`.

## Tests

- Unit: `taskParamsForKind` for both kinds (ec2Tags/label/static/mixed-empty-label/empty-peers/nil-target; RestartPod params + UID threading), `restart-pod` execution (UID match/no-op/missing/terminating/multi-replica guard, **same-second replacement** regression), CEL union (reject zero/multiple payloads, kind immutability).
- Reconcile (fake client): both kinds happy-path, empty-peers→Failed, label-empty-resolved→Failed, pod-missing→waits→completes, never-Ready→Timeout.
- envtest: CEL validation against a real apiserver.
- `make test` + `make test-integration` green; `make lint` 0 new findings vs main.

## Follow-up (deferred)

- Promote peer-drift to an automatic controller trigger once proven.
- `seid` pod has no explicit `terminationGracePeriodSeconds` (inherits 30s) — if seid needs longer to flush on shutdown, worth setting in the StatefulSet generator. Independent of this PR.


internal/controller/nodetask/controller.go
```diff
@@ -289,6 +344,34 @@ func (r *SeiNodeTaskReconciler) driveTask(ctx context.Context, cr *seiv1alpha1.S
...
+// Per-kind default timeouts applied when spec.timeoutSeconds is 0 (unset).
+// These bound kinds whose completion depends on a pod becoming Ready or a quick
+// disk write — without a bound, a wedged seid (e.g. DiscoverPeers wrote peers
+// that crash-loop the node) would requeue forever with no Failed transition.
+// Sidecar-backed gov/await kinds keep the unbounded default: they poll the
+// chain and an operator chooses the bound via spec.timeoutSeconds.
+const (
+	defaultRestartPodTimeout    = 10 * time.Minute
+	defaultDiscoverPeersTimeout = 2 * time.Minute
+)
```
bdchatham: Please put at the top of the file.

internal/controller/nodetask/controller.go
```diff
@@ -374,10 +494,58 @@ func taskParamsForKind(cr *seiv1alpha1.SeiNodeTask, target *seiv1alpha1.SeiNode)
...
+		return "", nil, fmt.Errorf("%w: kind %q is not wired in this build", errUnsupportedKind, cr.Spec.Kind)
 	}
 }
 
+// errUnsupportedKind sentinels the default case in taskParamsForKind: a kind
+// the CRD enum admits but this build does not wire. The synthesis site routes
+// it to reason=UnsupportedKind; every other (payload/param) failure is
+// ParamsBuildFailed. Reasons are a public enum (CLAUDE.md), so UnsupportedKind
+// must mean only that.
+var errUnsupportedKind = errors.New("unsupported kind")
```
bdchatham: top of file

internal/controller/nodetask/controller.go
```diff
@@ -374,10 +494,58 @@ func taskParamsForKind(cr *seiv1alpha1.SeiNodeTask, target *seiv1alpha1.SeiNode)
...
+				continue
+			}
+			sources = append(sources, sidecar.PeerSource{
+				Type:      sidecar.PeerSourceStatic,
+				Addresses: target.Status.ResolvedPeers,
+			})
+		}
+	}
+	return sources
+}
```
bdchatham: Why does this need a new implementation? We have peering config handling in the other controllers. Can't it be shared?

internal/controller/nodetask/controller.go
bdchatham, line 344: I would suggest moving the actual task param creation out of this method. This method simply determines which one to call. The param creation can come from another method. That other method should probably live in the task package/module

internal/controller/nodetask/controller.go
bdchatham: Done — moved `defaultRestartPodTimeout`/`defaultDiscoverPeersTimeout` into the top-of-file `const(` block (2a91ca9).

internal/controller/nodetask/controller.go
bdchatham: Done — the unsupported-kind sentinel moved to the `task` package as exported `task.ErrUnsupportedKind` (its only consumer is now there, since param construction moved per your other comment); consumer uses `errors.Is(err, task.ErrUnsupportedKind)` (2a91ca9).

internal/controller/nodetask/controller.go
bdchatham: Good call — removed the duplicate. There's now a single `task.DiscoverPeerSources(node)`; `planner.discoverPeersTask` is reduced to a one-liner wrapping it. Bonus: the shared empty-label-skip also removed a latent planner edge that could write an empty `persistent-peers` for a not-yet-resolved label source (2a91ca9).

internal/controller/nodetask/controller.go
bdchatham: Done — `taskParamsForKind` is now a thin dispatcher (~12 lines) that just resolves the target and calls `task.SeiNodeTaskParamsFor`. All seven per-kind param constructors moved to `internal/task/seinodetask_params.go`; behavior-preserving for the existing kinds (2a91ca9).

bdchatham:
Pushed `2a91ca9` addressing all review feedback:
- **Cursor Bugbot (HIGH, timeout includes target wait):** execution budget now measured from a new `status.task.executionStartedAt` (stamped once the target is ready), so the requirePhase wait is no longer charged against the per-kind timeout. `spec.timeoutSeconds` is now execution-only (documented).
- **CI lint (goconst):** repeated test literals hoisted to consts.
- **Your 4 comments:** consts → top of file; `ErrUnsupportedKind` → task package; shared peer-source builder (planner de-duped); `taskParamsForKind` → thin dispatcher with per-kind constructors in the task package; verbose doc comments trimmed.

Two expert cross-review rounds + a regression pass signed off on the refactor. Note: `spec.timeoutSeconds` semantics changed from total→execution-only (additive `executionStartedAt` status field, back-compat).

internal/controller/nodetask/controller.go
bdchatham: Resolved in 2a91ca9 — the execution budget now measures from a new `status.task.executionStartedAt` (stamped once the target is ready), so the requirePhase wait is no longer charged against it. The re-anchored HIGH inline is from the original commit (5244a92); Bugbot's latest review on 2a91ca9 didn't re-raise it.

internal/planner/planner.go
```diff
@@ -657,33 +657,7 @@ func snapshotRestoreTask(snap *seiv1alpha1.SnapshotSource) sidecar.SnapshotResto
...
-			// ResolvedPeers is pre-composed `<node_id>@<host>:<port>`;
-			// route as static so the sidecar writes them verbatim.
-			sources = append(sources, sidecar.PeerSource{
-				Type:      sidecar.PeerSourceStatic,
-				Addresses: node.Status.ResolvedPeers,
-			})
-		}
-	}
-	return sidecar.DiscoverPeersTask{Sources: sources}
+	return sidecar.DiscoverPeersTask{Sources: task.DiscoverPeerSources(node)}
```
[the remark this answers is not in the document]
bdchatham: Fixed in 1aaaefa. The planner no longer submits a zero-source discover-peers: a new controller-side `discover-peers-init` task re-derives sources from the live node each reconcile and, when sources are empty (label not yet resolved), defers via Failed→retry-budget (bounded by discoverPeersMaxRetries) instead of writing — so it never wipes persistent-peers and converges once status.resolvedPeers populates. The nodetask path keeps its synthesis-time fail-fast.

bdchatham:
Pushed `1aaaefa`:
- **Bugbot MEDIUM (init plan empty discover-peers):** fixed — new controller-side `discover-peers-init` task re-derives peer sources from the live node each reconcile; empty/unresolved sources defer via the retry budget (bounded) instead of submitting a zero-source write that could wipe `persistent-peers`. Converges once `status.resolvedPeers` populates.
- **Bugbot HIGH (timeout includes target wait):** already resolved in `2a91ca9` (execution budget from `executionStartedAt`); the re-anchored inline was the original-commit comment.
- **DRY (maintainer follow-up):** extracted a shared `podCycle` helper so `restart-pod` reuses `replace-pod`'s pod-deletion plumbing — both behavior-preserving.

Regression cross-review signed off on both. CI + Bugbot re-running.

internal/controller/nodetask/controller.go
```diff
@@ -132,16 +138,34 @@ func (r *SeiNodeTaskReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 			}
+			// Capture the target pod's UID once, at synthesis, to
+			// content-address the restart (see SeiNodeTaskExecution.RestartedPodUID).
+			if cr.Spec.Kind == seiv1alpha1.SeiNodeTaskKindRestartPod {
+				uid, err := r.targetPodUID(ctx, target)
+				if err != nil {
+					return r.flush(ctx, cr, before, statusBase, observedPhase, ctrl.Result{}, err)
+				}
+				exec.RestartedPodUID = string(uid)
+			}
```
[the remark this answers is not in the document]
bdchatham: Fixed in f477e34. RestartPod now never completes without restarting: if no owned pod is observed at synthesis it defers (requeue, bounded by requirePhaseTimeout → Failed/RestartTargetUnresolved) until a real pod UID is captured; and as a backstop the task's Status stays Running on an empty UID (never completes on an arbitrary Ready pod). The pod-exists path is unchanged — captures the UID and requires a different-UID Ready pod.

internal/controller/nodetask/controller.go
```diff
@@ -132,15 +138,46 @@ func (r *SeiNodeTaskReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	// kind/payload mapping first so we fail-fast on a malformed spec before
+	// stamping anything. UnsupportedKind is reserved for a genuinely unwired
+	// kind (the default case); a payload/param-build failure for a wired kind
+	// surfaces as ParamsBuildFailed, matching driveTask.
 	if fatal == nil && cr.Status.Task == nil {
 		if _, _, err := taskParamsForKind(cr, nil); err != nil {
-			r.markFailed(cr, now, "UnsupportedKind", err.Error())
+			reason := "ParamsBuildFailed"
+			if errors.Is(err, task.ErrUnsupportedKind) {
+				reason = "UnsupportedKind"
```
bdchatham: Make unsupported kind a field of ErrUnsupportedKind

internal/controller/nodetask/controller.go
```diff
@@ -132,15 +138,46 @@ func (r *SeiNodeTaskReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+			}
+			r.markFailed(cr, now, reason, err.Error())
 		} else {
+			// RestartPod must content-address a REAL pod (see
+			// SeiNodeTaskExecution.RestartedPodUID). Capturing an empty UID would
+			// make restart-pod a silent no-op that completes on the first owned
+			// Ready pod — including a pod that was never restarted. When no owned
+			// pod is observed yet (cache lag, or a brief gap between the target
+			// reaching Running and its pod existing), defer synthesis and re-attempt
+			// the capture on the next reconcile, bounded so it can't spin forever.
```
bdchatham: If none if found, we should just return an error with a proper reason.

bdchatham:
## Ready for human review — automated bar met (`f477e34`)

All CI checks green and **Cursor Bugbot passes** with no new findings:
```
Cursor Bugbot   pass     lint   pass     test   pass     test-integration   pass     verify-generated   pass
```

Review trail (each fix had a regression cross-review before push):
| Finding | Resolution |
|---|---|
| Bugbot HIGH — timeout includes requirePhase wait | `executionStartedAt` (exec-only budget) — `2a91ca9` |
| CI lint goconst | hoisted test literals to consts — `2a91ca9` |
| Maintainer ×4 (consts→top, sentinel→task pkg, shared peer builder, thin dispatcher) | `2a91ca9` |
| Bugbot MEDIUM — init zero-source discover-peers | `discover-peers-init` defers via retry budget — `1aaaefa` |
| Maintainer — DRY restart/replace | shared `podCycle` helper — `1aaaefa` |
| Bugbot MEDIUM — empty-UID silent skip | synthesis-defer + task backstop — `f477e34` |

The three inline comments still shown are all from earlier commits and resolved above (the HIGH @387 is the original first-commit comment re-anchored to the diff; Bugbot's per-commit check is green).

Note: `spec.timeoutSeconds` is now **execution-only** (additive `status.task.executionStartedAt`, back-compat). Deferred follow-up (separate): promote peer-drift to an automatic controller trigger once proven.

Over to you for the final look, @bdchatham.

internal/controller/nodetask/controller.go
bdchatham: Done in 39bd88c — `ErrUnsupportedKind` is now a typed error `type ErrUnsupportedKind struct { Kind SeiNodeTaskKind }` (carries the kind); the default dispatch case returns `&ErrUnsupportedKind{Kind: cr.Spec.Kind}` and the controller detects it via `errors.As`.

internal/controller/nodetask/controller.go
bdchatham: Done in 39bd88c — replaced the bounded-defer (`awaitRestartTarget`) with a fail-fast: no owned pod at synthesis → `Failed`/`RestartTargetPodNotFound`. `targetPodUID` now reads uncached (APIReader) so cache lag isn't a false negative, and the task-side backstop (empty UID never completes) stays as defense-in-depth.

bdchatham:
Pushed `39bd88c` — unwound the over-built complexity per your call + addressed the two new comments:
- **Unwind:** deleted `discover-peers-init`; reverted the planner's init discover-peers to **main** (byte-identical — `git diff origin/main -- internal/planner/planner.go` is empty); the nodetask `DiscoverPeers` now builds its **own** peer sources with its own empty/label-unresolved fail-fast (slight, deliberate duplication — no shared interface to conflict). The nodetask still can't submit a zero-source write that would wipe `persistent-peers`.
- **:149** — `ErrUnsupportedKind` is now a typed error carrying the kind (`errors.As`).
- **:159** — RestartPod no-pod-at-synthesis → fail fast (`RestartTargetPodNotFound`), dropping the `awaitRestartTarget` defer; uncached read avoids cache-lag false negatives; task-side backstop kept.

Net: the init path is untouched vs main, and the two new tasks are self-contained. Expert regression cross-review signed off. CI + Bugbot re-running.

bdchatham:
Pushed `ec2b591`. Status of the Bugbot findings on the latest commit:
- **In-flight tasks lose timeout (HIGH)** — **fixed**: `driveTask` lazy-stamps `executionStartedAt=now` when a populated `status.task` has a nil anchor (a task synthesized by a pre-this-change controller), so in-flight legacy tasks get a bounded fresh budget on upgrade instead of running unbounded. Idempotent; not anchored to `startedAt` (would back-date the deadline).
- **Task timeout includes target wait (HIGH)** — **stale/re-anchored**, resolved in 2a91ca9: the budget measures from `executionStartedAt` (stamped after the target is ready), not `startedAt`.
- **Empty pod UID skips restart (MEDIUM)** — **resolved** in 39bd88c: no owned pod at synthesis → fail fast (`RestartTargetPodNotFound`); task-side backstop keeps an empty UID in Running (never completes). The defer machinery is gone.

CI + Bugbot re-running on ec2b591.

internal/controller/nodetask/controller.go
```diff
@@ -132,15 +145,46 @@ func (r *SeiNodeTaskReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	// kind (the default case); a payload/param-build failure for a wired kind
+	// surfaces as ParamsBuildFailed, matching driveTask.
 	if fatal == nil && cr.Status.Task == nil {
 		if _, _, err := taskParamsForKind(cr, nil); err != nil {
-			r.markFailed(cr, now, "UnsupportedKind", err.Error())
+			reason := "ParamsBuildFailed"
+			var unsupported *task.ErrUnsupportedKind
+			if errors.As(err, &unsupported) {
+				reason = "UnsupportedKind"
+			}
```
bdchatham: give these errors a reason so that you don't need an if-statement. it defeats the purpose to include it just on one.

internal/controller/nodetask/controller.go
```diff
@@ -132,15 +145,46 @@ func (r *SeiNodeTaskReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+			execStart := metav1.NewTime(now)
 			cr.Status.Task = &seiv1alpha1.SeiNodeTaskExecution{
-				ID:     task.DeterministicTaskID(string(cr.UID), string(cr.Spec.Kind), 0),
-				Status: seiv1alpha1.TaskPending,
+				ID:                 task.DeterministicTaskID(string(cr.UID), string(cr.Spec.Kind), 0),
+				Status:             seiv1alpha1.TaskPending,
+				ExecutionStartedAt: &execStart,
+				RestartedPodUID:    restartedPodUID,
 			}
```
bdchatham: Change the api to take a pod UID so we don't need to fetch it.

PR #417, opened 2026-06-16
refactor(api): SeiNetwork clean-break — genesis-required Kind, delete SeiNodeDeployment

bdchatham:
## What

Replaces `SeiNodeDeployment` with a new genesis-**required** `SeiNetwork` Kind (`sei.io/v1alpha1`, `shortName: sn`) and deletes the SeiNodeDeployment controller in the same change — one Kind, no coexistence. Networking is deserted entirely (GitOps owns it since PLT-451); the controller keeps the genesis ceremony, the validator-pool fan-out, and the internal ClusterIP.

`SeiNetwork` models exactly a chain's **genesis validator pool**. Followers (rpc/archive/full) are not modeled here — they become individual `SeiNode`s provisioned by the skills (a later milestone).

## Why

Genesis was an optional field bolted onto a homogeneous replica set, but the load-bearing use-case (self-service custom chains via `seictl` + `harbor-dev`) treats genesis as the product. This makes the Kind's existence assert "a genesis-bootstrapped network." Full rationale, one-way doors, and sequencing: [`seinetwork-lld.md`](https://github.com/sei-protocol/bdchatham-designs/blob/main/designs/seinetwork/seinetwork-lld.md).

## ⚠ Deploy gate — do not deploy this image to a live cell yet

This deletes the SND controller, so the resulting image **must not** be deployed to any cell that still has live `SeiNodeDeployment` CRs (arctic-1, atlantic-2, pacific-1). The live fleet is migrated off SND onto `SeiNetwork` + standalone `SeiNode`s **first** (a prerequisite milestone); pacific-1 mainnet is last and held. Merging to `main` is safe — it builds the image, it does not deploy it.

## Review

Cohort cross-reviewed post-cleanup: kubernetes / sei-network / security / idiomatic all APPROVE. Trust invariants preserved (double-sign CEL, no Secret RBAC, distinctness lattice inherited via the unchanged SeiNode, node_id continuity). The frozen `sei.io/nodedeployment` label key is retained so GitOps selectors keep matching.

## Follow-ups (tracked, not in this PR)

- Controller-side genesisHash plumbing: write `status.genesisHash` from the in-band assemble result and pass it into followers' `ConfigureGenesisTask.ExpectedGenesisHash` (consumes the matching `seictl` change; required before the deploy gate opens).
- Update this repo's `CLAUDE.md` to the SeiNetwork reality.
- M2/M3: reshape `seictl` (genesis + follower paths) and `harbor-dev`.

## CI notes

`golangci-lint` and envtest (`test-integration`) couldn't run in the authoring env (toolchain pin / missing setup-envtest binaries); they must go green here. The runner may need `GOTOOLCHAIN=go1.26.0`.


internal/controller/seinetwork/labels.go
```diff
@@ -0,0 +1,85 @@
...
+	"strconv"
+
+	seiv1alpha1 "github.com/sei-protocol/sei-k8s-controller/api/v1alpha1"
+)
+
+const (
+	// groupLabel is frozen as a GitOps networking selector key; child Services
+	// and listing select on it.
+	groupLabel          = "sei.io/nodedeployment"
+	groupOrdinalLabel   = "sei.io/nodedeployment-ordinal"
```
bdchatham: This needs to be updated.

bdchatham:
**Bugbot findings addressed in `9634080`** (+ owner-directed model changes):

- **NodesReady not seeded** (Med) — fixed. `seedAlwaysPresentConditions` now seeds `NodesReady=False/Pending` and `RolloutInProgress=False/AllUpToDate`; `updateStatus` computes the live values after. Full always-present set is now guaranteed on early-return paths.
- **DataVolume not propagated on update** (Med) — resolved by making `dataVolume` **CEL-immutable**. A StatefulSet's `volumeClaimTemplates` are immutable, so a post-create change could never take effect; we now reject it at admission rather than silently ignore it.
- **Scale-up blocked during genesis** (Med) — resolved by making `replicas` **CEL-immutable**. The genesis validator count is fixed at the ceremony (you can't add a validator to an already-minted genesis without on-chain registration, which is a follower/standalone concern, not a SeiNetwork).
- **Overrides diff ignores nil maps** (Low) — **false positive.** `ensureSeiNode` uses `maps.Equal`, which is length-based: `maps.Equal(nil, map[string]string{})` returns `true`. Nil and empty are already treated equal, so there's no spurious reconcile update. Left unchanged.
- **Alert references missing metric** (Low) — fixed. `sei_controller_seinetwork_reconcile_substep_duration_seconds_bucket` was never registered (the controller emits only `phase`/`replicas` OTel instruments). Repointed `ControllerHighReconcileLatency` to the controller-runtime built-in `controller_runtime_reconcile_time_seconds_bucket`, which the registry does export.
- **`sei.io/*` label stamping rule** — the new canonical `sei.io/seinetwork{,-ordinal}` labels are stamped inside `seiNodeLabels` alongside the frozen `sei.io/nodedeployment{,-ordinal}` keys (kept for GitOps-selector continuity; selectors migrate to the new keys later, then the old ones retire).

Verified: `golangci-lint` 0, `make build`/`make test`, and `make test-integration` (envtest 1.34.0) all pass.

bdchatham:
**Bugbot disposition (commit `192ad9b`)** — all threads resolved:

*Confirmed addressed (no code change needed):*
- **DataVolume not propagated** — `dataVolume` is CEL-immutable (`seinetwork_types.go:29`, CRD rule `self.dataVolume == oldSelf.dataVolume`); a volumeClaimTemplate edit is rejected at admission, so there's no update path to sync.
- **NodesReady not seeded** — seeded `False/Pending` in `seedAlwaysPresentConditions` (`status.go:157-158`).
- **Alert references missing metric** — repointed to the registered `controller_runtime_reconcile_time_seconds_bucket`.
- **Overrides diff ignores nil maps** — false positive: `nodes.go:154` uses `maps.Equal`, which is length-based (nil == empty).
- **Scale-up blocked during genesis** — `replicas` is CEL-immutable (`self.replicas == oldSelf.replicas`); the genesis validator set is fixed at the ceremony, rejected at admission.

*Fixed in `192ad9b`:*
- **Missing role label on children** — `seiNodeLabels` now stamps `sei.io/role=validator` (a SeiNetwork is always a validator pool); covered by `TestChild_RoleLabelStamped`.
- **Peer docs claim missing revision label** — `LabelPeerSource` doc corrected (revision label removed with the rollout state machine; now documents `sei.io/seinetwork{,-ordinal}` + `sei.io/role`).
- **Sidecar resources not propagated** — `ensureSeiNode` now diffs the whole `Sidecar` struct via `equality.Semantic.DeepEqual` (incl. `Resources`); covered by `TestChild_SidecarResourcesPropagate`.

All gates green: golangci-lint 0, build, `make test`, and `make test-integration` (envtest 1.34.0).

Issue #175, opened 2026-05-05
feat(api): split api/ into a Go submodule for cross-repo typed import


bdchatham:
## Problem

The typed CRD surface at `api/v1alpha1/` (`SeiNode`, `SeiNodeDeployment`, etc.) lives inside the main `github.com/sei-protocol/sei-k8s-controller` module. Downstream consumers that want typed access (seictl's new `nodedeployment` CLI per `sei-protocol/seictl/docs/design/nodedeployment-cli.md`, future controllers, the seiload harness) cannot import these types directly: `sei-k8s-controller/internal/...` imports `github.com/sei-protocol/seictl/sidecar/client` from 17+ sites, so a typed import in the reverse direction creates a Go module-level cycle.

The seictl `nodedeployment` CLI implementation hits this immediately. The architectural fallback is `unstructured.Unstructured` everywhere — verbose accessors, no compile-time type safety, ongoing drift risk against schema changes — paid by every consumer that wants typed CRD access.

## Impact

- seictl's PR-1 for the new CLI needs to choose: either ship with `unstructured` and refactor later, or block on the split landing here.
- Future consumers (additional controllers, fuzzers, upcoming harnesses) each pay the same `unstructured` tax until the split lands.
- "Single source of truth for CRD types" is the standard K8s ecosystem pattern. Cluster-API, Argo CD, Longhorn, Karpenter, and Crossplane all split `api/` into a leaf module for exactly this reason.

## Relevant experts

- `kubernetes-specialist` — kubebuilder/controller-runtime conventions, module structure.

## Proposed approach

Extract `api/` into a Go submodule (`github.com/sei-protocol/sei-k8s-controller/api`):

1. Add `api/go.mod` declaring the new module path. Direct dependencies should be limited to `k8s.io/apimachinery`, `k8s.io/api`, and `sigs.k8s.io/controller-runtime/pkg/scheme` — no project-internal imports.
2. Verify `api/v1alpha1/` is leaf-clean (no imports from `sei-k8s-controller/internal/...` or `cmd/...`). If not, lift the offending shared code into `api/` or push it down into `internal/`.
3. Update the main module's `go.mod` to require the api submodule. Use a `replace` directive pointing at `./api` for local development; `go work` syncs across both.
[...]

PR #421, opened 2026-06-20
feat(sdk): Go SDK for in-process SeiNetwork/SeiNode orchestration (WS-E)

bdchatham:
## What

Lands the **WS-E integration refactor** — a Go SDK under `sdk/` — so CICD chaos/test harnesses can provision a genesis **SeiNetwork** + a follower **SeiNode** fleet, wait for readiness, read endpoints, and tear down **in one Go program** (replacing the bash in `platform`'s `k8s_nightly`).

## API (one-way-door surface)

- `sei.Open(ctx, name) (*Client, error)` — `database/sql`-style; provider registry with **blank-import flavor selection** (`_ ".../sdk/sei/provider/k8s"`), env precedence (explicit → `SEI_PROVIDER` → presence: `SEI_NODE_CLUSTER`⇒k8s / `SEI_LOCAL`⇒local; both-set fail-fast).
- `ProvisionNetwork` / `ProvisionFleet` / typed `Endpoints()` (per-pod, read verbatim from `.status.endpoint` — never reconstructed) / idempotent `Teardown`. `Class` error enum + `IsTimeout`/`IsFailed`.
- **Canonical readiness probe** (the SDK is now the single source of truth): the Sei CometBFT unwrapped-envelope `/status` decode + the consensus-honest gate (`height>1 AND catching_up==false`) then `eth_blockNumber`. Canonical `sei.io/role=node` / `sei.io/seinetwork` label constants.
- `provider/k8s` (SSA + label stamping + `.status.endpoint` reads + serial fan-out); `provider/local` (registered stub); `cmd/sei` thin `up` shell that dogfoods the API.

## Why in-module (not a standalone repo)

Per the decision to land in an existing repo: importing `api/v1alpha1` is now **in-module** — no cross-module version skew, no `replace`. `seitask` lives here too, so its convergence onto the canonical probe is a trivial in-module import (follow-up issue). External importers (seictl/harnesses) pull the full controller graph — the accepted tradeoff; the `api/` leaf-module split (#175) would lighten that later.

## Validation

- `go build ./...`, `go vet ./...`, `go test ./... -race`, `golangci-lint run ./sdk/...` (0 issues), `gofmt -s -l sdk/` (clean), and `make test` over the whole module — all green.
- SSA via `ctrlclient.Apply` intentionally matches the controller's `internal/task` pattern (one `//nolint:staticcheck` SA1019 with that rationale; module-wide migration tracked separately).
- `.golangci.yml`: extended the existing `internal/*`/`api/*` `lll`/`dupl` exclusion to `sdk/*` (peer application code).

## Provenance

WS-E LLD signed off (xreview RESOLVED R1→R2; SDK-canonical-probe decision); implementation idiom-reviewed clean (the cross-namespace peer-wiring bug it caught is fixed + regression-tested). Convergence follow-ups (seitask imports the canonical probe; `#175`; the provider-key drift guard) filed separately.


bdchatham:
Bugbot's last finding ("Loop defer deletes wrong nodes", example_test.go:55) is a **false positive**: `node` is declared with `:=` *inside* the loop body (line 47), so it's a fresh variable each iteration — each `defer` closes over its own `node` and all RPC nodes are deleted correctly. This is distinct from the loop-*variable*-capture bug (which is about `i`), and the module is go 1.26 where loop variables are per-iteration anyway. No change needed.

Merging: the reshaped thin-CRUD SDK is design-confirmed, idiom-reviewed clean (zero findings), and CI-green; the earlier (real) Bugbot findings were all in the orchestration complexity that this reshape removed.

PR #428, opened 2026-06-22
feat(test): WS-I sei integration suite — TestBenchmark provision spine [DRAFT]

bdchatham:
## WS-I Step 1 (foundation) — Go-native nightly harness as `go test` targets

First increment of the Go-native test harness that replaces the Chaos-Mesh Workflow DAG + seitask Task pods + workflow-vars ConfigMap. **Draft** — lands the architecture + provisioning spine; the seiload drive + S3 report is the next increment (marked TODO + `t.Skip`).

### What's here
- `test/integration/harness_test.go` — shared machinery: `spec`/`chain` (local Go state, replacing workflow-vars), `provision` (SDK in-process: CreateNetwork 4 validators + N RPC SeiNodes, each waited Running→caught-up→EVM-serving), `teardown` via `t.Cleanup`, `sei.io/harness-run` label const, env gate.
- `test/integration/benchmark_test.go` — `TestBenchmark` (load suite).

### Architecture decisions (WS-I LLD)
- **D3** Go orchestrator, SDK in-process; seiload + chaos are decoupled units the suite *applies*, not constructs.
- **D4** plain `go test` targets in `_test.go` (not a CLI binary) — Go's `_test.go` rule guarantees **zero test code in any production binary**; `//go:build integration` hides it from default CI. Verified: `go build ./...` clean, no integration code in `cmd/` deps.
- **D5** in-cluster CronJob vehicle; one `go test -c` image, one CronJob per target (`-test.run TestX`). Targets: TestBenchmark / TestChaosSuite / TestChainUpgrade / TestRelease.
- **Stages wholesale seitask-runner removal:** imports only `sdk/sei` (+ k8s provider).

### Verification
`gofmt` clean · `go build ./...` clean · `go vet -tags integration ./test/integration` clean · `go test -c -tags integration` → binary exposes `TestBenchmark` · skips cleanly without `SEI_NODE_CLUSTER`.

### Next increment
seiload as a decoupled unit (apply its own manifest w/ `evmEndpoints()`, stamped `harness-run`) → wait → read S3 report → assert TPS. Then mark ready + Coral/Bugbot/CI gate.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #429, opened 2026-06-23
fix(sdk,test): cascade-delete ephemeral chain validators (Bugbot high-sev follow-up to #428)

bdchatham:
## Follow-up to #428 — Bugbot high-severity finding

Bugbot flagged on #428 (verified valid): the integration harness's `teardown` **orphans validator SeiNodes**.

### Root cause (confirmed in-tree)
- `provision` creates the genesis `SeiNetwork`; `teardown` deletes it + the RPC `SeiNode`s.
- The CRD defaults `DeletionPolicy=Retain` (`+kubebuilder:default=Retain`, `seinetwork_types.go:98`), and under Retain the controller **strips the validator children's ownerRef** (`removeOwnerRef`, `internal/controller/seinetwork/nodes.go:301`) so they survive network deletion.
- The validators are **controller-created** → they never carry `sei.io/harness-run` → neither `t.Cleanup` nor the label-GC sweep reaps them. **Leaks 4 validators + PVCs every run** in the shared `nightly` namespace.

### Fix
- Add `DeletionPolicy` to the SDK `NetworkSpec` (plain `string` + `DeletionDelete`/`DeletionRetain` constants — core stays stdlib-only; threaded into `renderNetwork`). Same additive one-way-door pattern as `Labels` (#428).
- The harness sets `DeletionDelete` — an ephemeral chain cascade-deletes its validators (+ PVCs) on teardown. Correct for a throwaway test chain (distinct from production fleet nodes, which stay Retain).
- Locked by a `render_test` assertion that `DeletionPolicy` threads through.

### Verification
`go build ./...` clean · `go test ./sdk/...` pass (incl. new assertion) · `go test -c -tags integration` compiles · golangci-lint 0 issues.

### DIFF
--- a/sdk/sei/provider/k8s/render.go
+++ b/sdk/sei/provider/k8s/render.go
@@ -37,6 +37,9 @@ func renderNetwork(spec sei.NetworkSpec, namespace string) *seiv1alpha1.SeiNetwo
 		Spec: seiv1alpha1.SeiNetworkSpec{
 			Image:    spec.Image,
 			Replicas: int32(spec.Validators),
+			// "" leaves the CRD default (Retain); a caller sets Delete so an
+			// ephemeral chain's validators cascade-delete instead of orphaning.
+			DeletionPolicy: seiv1alpha1.DeletionPolicy(spec.DeletionPolicy),
 			Genesis: seiv1alpha1.GenesisCeremonyConfig{
 				ChainID: spec.Name, // chain ID defaults to the network name
 			},
--- a/sdk/sei/provider/k8s/render_test.go
+++ b/sdk/sei/provider/k8s/render_test.go
@@ -9,10 +9,11 @@ import (
 func TestRenderNetwork_ChainIDDefaultsToName(t *testing.T) {
 	spec := sei.NetworkSpec{
 		Name: testNet, Image: testImage, Validators: 4,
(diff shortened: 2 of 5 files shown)


bdchatham:
## Comment-standards expert review applied (idiom D10 + prose register)

Per request, ran a two-lens review scoped to comment register across the WS-I harness + SDK surface. Both converged; all findings applied in `d9d6b00`:
- **stripped meta/ID cruft** from code comments (`Bugbot high-sev`, `WS-I step N`, `decision D*`) — that context lives here in the PR, not in source
- **dropped migration-history framing** (`replaces…`, `used to pass…`) → present-state only
- **fixed one drift** (gating): `TestBenchmark`'s doc claimed seiload drive/report the body `t.Skip`s
- **trimmed agent-verbose** package doc + de-duplicated the `DeletionPolicy` rationale to one canonical home

Vetted-clean and kept: the `// -> spec.X` field mappings, blank-import rationale, SIGTERM/why comments — the standard is why-not-what + crisp, not fewer comments.

Durable follow-up the idiom lens recommended: graduate this into `/pr-quality` as a grep gate (`Bugbot|WS-I|xreview|decision D[0-9]` in `*.go` comments outside `TODO(`) so it's mechanically enforced. Will forward to Tide.

PR #430, opened 2026-06-23
feat(test): TestBenchmark drives seiload (WS-I load suite)

bdchatham:
## WS-I — complete the load suite (TestBenchmark)

Builds on the merged foundation (#428) + SDK `Labels`/`DeletionPolicy` (#429). After provisioning the chain + RPC fleet, `TestBenchmark` now drives seiload and asserts the chain survived the load.

### Flow
1. Render the platform seiload profile (read from the `seiload-profiles` ConfigMap) with the per-run chain id + the fleet's EVM endpoints; write a per-run profile CM stamped `sei.io/harness-run`.
2. Apply **seiload's own Job manifest** (embedded template, parameterized — its Job spec is *not* constructed in Go; D3 decoupling) carrying the metrics scrape label + `harness-run`.
3. Wait for the Job to run the full load to completion.
4. Assert the chain stayed live (post-load `WaitCaughtUp` on node-0).

### Assert model (deliberate)
Pass/fail = **Job completion + post-load chain liveness**. A throughput/regression gate belongs in **telemetry** — a PromQL query over the run's metrics (the conventional path, like validate-release reads Grafana) — *not* parsed from the report. Left as a documented follow-up; the Job already carries the metrics scrape label so it's a drop-in.

### Decoupling / cleanup
- seiload runs from its own manifest; the profile is platform-owned (read from the cluster CM, not vendored).
- Per-run profile CM + seiload Job carry `sei.io/harness-run`; `t.Cleanup` deletes them on normal exit, the label-GC sweep backstops abnormal exit.

### Verification
`gofmt` clean · `go build ./...` clean (no integration code in prod deps) · `go vet -tags integration` clean · golangci-lint 0 issues · `go test -c -tags integration` → `TestBenchmark` · skips without `SEI_NODE_CLUSTER`.

### Next
Live smoke on harbor (demonstration) + platform PR to point the load CronJob at the new image and update the podMonitor selector to `sei.io/harness-run`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #431, opened 2026-06-23
fix(test): provision chains with working seid config (smoke-proven)

bdchatham:
## Follow-up to #430 — make TestBenchmark actually bring up a chain

In-cluster smoke of TestBenchmark on harbor (run via an in-cluster pod with a harness SA) caught two real gaps the SDK's *generic* provisioning left vs the old scenario templates:

1. **Validators CrashLooped** — `seid` exits with `invalid state-commit.sc-write-mode "cosmos_only"`. Every scenario template set `storage.state_commit/state_store.write_mode=memiavl_only`; the SDK's `CreateNetwork` omitted it. `provision` now applies that **storage baseline** to the network + every node.
2. **Load followers** need the EVM tuning the rpc template carried (`evm.worker_pool_size/worker_queue_size/max_tx_pool_txs`); `TestBenchmark` passes it via `spec.rpcConfig`, merged over the baseline.

### Proof (live, harbor)
```
network ... ready
rpc-0: running -> caught up -> EVM serving
rpc-1: running -> caught up -> EVM serving
--- PASS: TestBenchmark (353.16s)   # provision -> 2m seiload -> all followers caught-up post-load -> clean teardown
```
Resources fully reaped after (DeletionPolicy cascade + t.Cleanup).

### Run-model RBAC note (for the cutover deliverable)
The SDK provisions via **server-side apply**, so the harness ServiceAccount Role needs `patch`/`update` on `seinetworks`/`seinodes`, not just `create`. Folds into the harness-Role manifest.

### Verification
`go build ./...` clean · golangci-lint 0 issues · `go test -c -tags integration` compiles · **passes in-cluster end-to-end**.

### DIFF
--- a/test/integration/benchmark_test.go
+++ b/test/integration/benchmark_test.go
@@ -46,6 +46,7 @@ func TestBenchmark(t *testing.T) {
 		seiloadProfile: envOr("SEILOAD_PROFILE", "nightly_evm_transfer"),
 		seiloadCommit:  envOr("SEILOAD_COMMIT_ID", ""),
 		durationMin:    envInt(t, "DURATION_MINUTES", 10),
+		storageConfig:  memiavlStorageConfig,
 		// EVM tuning the followers need to absorb the load (matches the load
 		// scenario's rpc overrides).
 		rpcConfig: map[string]string{
--- a/test/integration/benchmark_test.go
+++ b/test/integration/benchmark_test.go
@@ -46,6 +46,13 @@ func TestBenchmark(t *testing.T) {
 		seiloadProfile: envOr("SEILOAD_PROFILE", "nightly_evm_transfer"),
 		seiloadCommit:  envOr("SEILOAD_COMMIT_ID", ""),
 		durationMin:    envInt(t, "DURATION_MINUTES", 10),
+		// EVM tuning the followers need to absorb the load (matches the load
+		// scenario's rpc overrides).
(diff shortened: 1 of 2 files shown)


bdchatham:
## /xreview — 2 lenses (idiom · k8s-dissenter), RESOLVED

**k8s [correctness]:** `baseStorageConfig` was globalized in `provision`, but `memiavl_only` is **not** universal — `major-upgrade`'s validator template deliberately omits `write_mode` (the storage/migration path is what that suite tests). Globalizing it would silently distort `TestChainUpgrade` once wired (latent today; only `TestBenchmark` is wired). **Fixed:** storage config is now **per-suite** (`spec.storageConfig`) — load/release set `memiavlStorageConfig`, upgrade leaves it nil → controller default. Byte-identical for `TestBenchmark` (smoke still valid).

**k8s confirmed correct:** config-key passthrough is verbatim (`Config` → CRD `configOverrides`, no drift); the SSA/RBAC note (Role needs `patch`/`update` — SDK uses server-side apply) is right and is a platform manifest deliverable; `mergeConfig` correct. The dropped template keys it flagged are release/upgrade-template, not load — load is faithful.

**idiom:** reads native; applied the two style nits (trim `mergeConfig` comment; `maps.Clone` for the network config copy). Shared-global concern vetted clear.

No one-way doors. Smoke-proven end-to-end on harbor.

PR #433, opened 2026-06-23
feat(test): port 4 more chaos faults (packet-loss, cpu/time/dns)

bdchatham:
## WS-I — chaos faults batch 2

Adds 4 faults on the proven network-partition flow (#432), covering the remaining duration-bearing KINDS:

| fault | kind | target | effect |
|---|---|---|---|
| packet-loss | NetworkChaos | one validator | 15% correlated loss |
| cpu-stress | StressChaos | one validator | CPU saturation |
| time-skew | TimeChaos | one validator | -30s clock skew |
| dns-chaos | DNSChaos | all validators | resolution errors |

All f=1-bounded (chain degrades, doesn't halt); the unfaulted follower observes forward progress via `WaitHeightAdvances`. Each is a `//go:embed` template + a `chaosScenarios` entry; the infra (gates, 0-target guard, lifecycle) is unchanged from #432.

**Excluded (follow-ups):** one-shot faults (pod-failure/container-kill — no `duration`, need `oneShot` handling); follower-targeted rpc-chaos (2 HTTPChaos CRs, different observer); mempool (no CR + PromQL telemetry gate).

### Verification
`go build ./...` clean · golangci-lint 0 issues · `go test -c -tags integration` compiles · in-cluster smoke of all 4 in progress.


bdchatham:
## /xreview — sei-network (per-kind fidelity), RESOLVED + smoke-proven

In-cluster smoke ran all 4; the review + smoke together split them cleanly:

| fault | verdict | smoke |
|---|---|---|
| packet-loss | COMPATIBLE (real fault, valid assert) | PASS (chain degraded-yet-advanced) |
| cpu-stress | COMPATIBLE | PASS |
| time-skew | COMPATIBLE (survives via median-vote dominance, not 'tolerance') | PASS (skewed validator outvoted) |
| **dns-chaos** | **MISMATCH — false-green** | PASS **vacuously** |

**dns-chaos dropped from this PR.** The patterns (`<chain>`/`<chain>-internal`/`<chain>-rpc`) don't match the per-pod peer FQDNs seid resolves (`<chain>-N-0.<chain>-N.<ns>.svc…`), and live MConnections don't re-resolve mid-fault — so it injects but doesn't perturb consensus, and the under-fault progress assert passes against an undisturbed chain. It's a rediscovery/recovery fault, not steady-state-liveness — needs a recovery-focused assert + FQDN-matching patterns. Deferred to the outlier follow-up rather than ship a vacuous test.

Also: reworded the time-skew comment to name the real mechanism (median-vote dominance). Namespace-parameterized DNS patterns were a fidelity *improvement* over the hardcoded-nightly original (noted, but doesn't rescue dns-chaos).

This PR ships **packet-loss + cpu-stress + time-skew** — all smoke-proven to genuinely perturb the chain while it stays live.

PR #435, opened 2026-06-23
feat(test): port one-shot kill faults (pod-failure, container-kill)

bdchatham:
## WS-I — chaos one-shot kill faults

Two PodChaos kill faults that need distinct handling: **no `duration`, no `AllRecovered`**. A `chaosScenario.oneShot` flag routes them:
- skip the self-expiry tripwire (one-shot kills carry no duration by design),
- under-fault window is fixed (the kill + StatefulSet/kubelet restart) rather than duration-derived,
- skip `gateRecovered` (nothing to recover at the chaos-mesh level) — recovery = the killed pod rejoining, covered by the post `WaitCaughtUp`.

| fault | action | effect |
|---|---|---|
| pod-failure | pod-kill | kills one validator pod; StatefulSet recreates it |
| container-kill | container-kill | kills the seid container; kubelet restarts in place |

Chain stays live on the other 3 (f=1); the follower advances through the kill. Chaos suite → **11/14**.

**Remaining:** rpc-chaos (HTTPChaos×2, follower-targeted), dns-chaos (recovery assert), mempool (PromQL/telemetry).

### Verification
build/lint clean · `go test -c -tags integration` compiles · in-cluster smoke of both in progress.

### DIFF
--- a/sdk/sei/readiness.go
+++ b/sdk/sei/readiness.go
@@ -89,18 +89,27 @@ func latestHeight(ctx context.Context, hc *http.Client, tmRPC string) (int64, bo
 // stalled node, which still reports catching_up == false at a frozen height.
 // hc may be nil. The caller's ctx bounds the wait (IsTimeout on a deadline).
 func WaitHeightAdvances(ctx context.Context, hc *http.Client, tmRPC string, delta int64) error {
-	target := int64(-1) // set to start+delta on the first successful read
-	return pollUntil(ctx, fmt.Sprintf("%s height +%d", tmRPC, delta), func(ctx context.Context) bool {
-		h, ok := latestHeight(ctx, hc, tmRPC)
-		if !ok {
-			return false
+	tick := time.NewTicker(probeInterval)
+	defer tick.Stop()
+	var start, last int64 = -1, -1 // baseline + most-recent height, for the timeout error
+	for {
+		if h, ok := latestHeight(ctx, hc, tmRPC); ok {
+			if start < 0 {
+				start = h
+			}
+			last = h
+			if h >= start+delta {
+				return nil
+			}
 		}
-		if target < 0 {
-			target = h + delta
-			return false
+		select {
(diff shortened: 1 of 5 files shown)


bdchatham:
## /xreview — full Coral slate (idiom · systems · k8s-dissenter · sei-network), RESOLVED

Per the all-experts-every-PR rule. The slate caught a real **false-green** before merge:

| # | Finding | Lens(es) | Resolution |
|---|---|---|---|
| **block** | **Recovery never asserted** — suite checked only the unfaulted follower, so a killed validator that CrashLoops forever (f=1 keeps survivors live) shipped green. The platform `verify-cleanup` polls every validator Ready; dropped on port. | dissenter (block) + sei-network + systems | `waitValidatorsReady` after every fault — poll `sei.io/nodedeployment` pods to Ready. Restores the recovery assertion (also closes the partition v0-rejoin gap). |
| C | one-shot quorum guard | systems | gateInjected asserts kills hit **exactly 1** target (a future edit to 2/4 halts the chain; ≥1 wouldn't catch it) |
| C | timeout observability | systems | `WaitHeightAdvances` reports `start→last` — stalled-chain vs tight-window |
| — | gate "race" (sei-network) vs "no race" (systems) | resolved → **systems** | records are on the durable CR status, AllInjected set after; smoke confirmed `injectedCount>0` fires for pod-kill. Single-read kept; comment corrected. |
| style | named window const; fixed the misleading 'caught-up covers recovery' comment | idiom/dissenter | applied |

**Conceded sound by all four:** GVR `podchaos`, gateInjected reliability, cleanup of a dead-target PodChaos, `mode:one` can't hit the follower, WAL/memIAVL crash recovery, the oneShot branch shape (`bool` not a premature interface). Idiom: clean.

Re-smoke (network-partition + pod-failure + container-kill) with the recovery gate in progress — validates the killed validator actually returns to Ready.

PR #436, opened 2026-06-23
feat(test): port dns-chaos with corrected peer-FQDN patterns

bdchatham:
## WS-I — dns-chaos (corrected)

Re-adds dns-chaos, **deferred in #433** because its patterns (`<chain>`/`<chain>-internal`/`<chain>-rpc`) matched no name seid resolves → vacuous. Corrected to `<chain>-*.<ns>.svc.cluster.local`, which matches the per-pod peer FQDNs validators actually resolve (`<chain>-N-0.<chain>-N.<ns>.svc…`).

With the `injectedTargets>0` gate confirming the DNS interceptor is actually installed on the validators, this is a legitimate **DNS-resilience** test: the chain keeps producing despite resolution failures on its own names (established MConnections don't re-resolve, so DNS error doesn't tear down live peering). The template comment is explicit that it asserts *resilience*, not active consensus perturbation — honest about what a green means.

Fits the standard flow (gateInjected + height-advance + recovery Ready-gate). Chaos suite → **12/14** (rpc-chaos + mempool/PromQL remain).

### Verification
build/lint clean · in-cluster smoke in progress (validates the corrected patterns inject + the chain stays resilient + recovers).

### DIFF
--- a/test/integration/chaos_test.go
+++ b/test/integration/chaos_test.go
@@ -46,6 +46,9 @@ var diskIOLatencyTmpl string
 //go:embed faults/byzantine.yaml.tmpl
 var byzantineTmpl string
 
+//go:embed faults/dns_chaos.yaml.tmpl
+var dnsChaosTmpl string
+
 //go:embed faults/pod_failure.yaml.tmpl
 var podFailureTmpl string
 
--- a/test/integration/chaossuite_test.go
+++ b/test/integration/chaossuite_test.go
@@ -29,6 +29,7 @@ const (
 	rTimeChaos    = "timechaos"
 	rIOChaos      = "iochaos"
 	rPodChaos     = "podchaos"
+	rDNSChaos     = "dnschaos"
 )
 
 // chaosScenario is one fault ported from the platform chaos suite: a name, the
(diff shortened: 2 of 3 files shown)


bdchatham:
Closing — full slate blocked this. sei-network traced the chaos-mesh matcher: the pattern `<chain>-*.<ns>.svc.cluster.local` is **illegal** (DNSChaos requires `*` at the end only; mid-string `*` is rejected at insert). The rule-push errors → Apply returns before writing records → AllInjected never fires → dns-chaos hangs at gateInjected then times out. **Confirmed by the in-cluster smoke** (provisioned + caught-up, then stuck — no 'fault injected'). The injectedTargets guard correctly surfaced it as a hard fail, not a false-green.

Even with the *legal* trailing form (`<chain>-*`), the dissenter's point stands: with established MConnections that don't re-resolve, the under-fault height-advance has no failing branch on the DNS-resilience dimension — it'd be a vacuous-green. A real dns-chaos needs: (1) legal trailing-`*` pattern matching the peer FQDNs, (2) a direct fault-effect assertion (a DNS query inside a faulted validator returns SERVFAIL during the window), and (3) ideally a forced re-dial (bounce a validator mid-fault) so resolution is actually exercised. Deferring to do it properly with that design, grouped with the other fault-effect-assert outliers (rpc-chaos, mempool).

PR #443, opened 2026-06-24
refactor(sdk): reframe sei SDK as a wire-client product (Phase A)

bdchatham:
## What

Make the `sdk/sei` pieces idiomatically express a settled design decision: the SDK is a **wire-client product + contract owner**. Its providers (k8s, docker) are WIRE clients — they speak RPC/CRDs to an already-running chain, so version-pinning them is fine. An **in-process ("local") provider is a category error for the SDK**: it would link sei-chain's Go source (the controller can't transitively build it; and it co-versions the harness with the chain). The in-process harness lives in **sei-chain**, satisfies these handle interfaces *by shape*, and is registered/constructed **consumer-side**. The SDK stops pretending to host "local".

## Changes

- **Delete** the `sdk/sei/provider/local` `ErrNotImplemented` stub package.
- **Drop** the built-in `local` mode + `SEI_LOCAL` from `resolveMode` (and the resolve/ambiguity error text). The registry stays generic — a consumer may still `provider.Register("local", …)` its own in-process adapter and `Open(ctx, "local")` explicitly.
- **Keep `docker`** untouched (legitimate wire stub).
- **Name the contract seam:** `Provider`/`*Handle` + `*Spec` + the registry are the stable, dependency-light contract — the seam a future leaf module would extract — and must stay free of controller-runtime/apimachinery/CRD imports. Documented in `provider.go` and `CLAUDE.md`. Not extracted now (deferred).
- **Add** a `go list -deps` import-discipline test that mechanically enforces that seam on the core `sei` package.
- **Fix** the drift test to pin built-in modes to exactly `{k8s, docker}`; refresh package/`Provider`/`Object()` docs and `sdk/CLAUDE.md`.

No `Provider`/handle interface **signatures** change. No contract sub-package/module extraction, no harness changes, no sei-chain changes (Phases B/C).

## Verification

`go build ./...`, `go test ./sdk/... -race`, `go vet ./sdk/...`, `gofmt -s -l sdk/`, `golangci-lint run ./sdk/...` (0 issues) — all green.

## For reviewers

- The drift test moved from "contains" to "exactly equals `{docker, k8s}`" — stricter, so an *extra* built-in mode also fails. Intentional: the SDK's built-in surface is now closed.
- `TestResolveMode_ExplicitBeatsEnv` now resolves a `"consumer-mode"` name, doubling as coverage that a consumer passes its own mode explicitly (the only path left for a non-wire provider).
- The new `imports_test.go` shells out to `go list`; it confirmed core has zero hits while the k8s provider pulls apimachinery, so it discriminates correctly.

## Out of scope (Phase A)

Interface-signature changes; contract sub-package/module extraction; integration harness (Phase B); sei-chain (Phase C).


bdchatham:
bugbot run

bdchatham:
## Review-gate: PASS — merging

**Coral slate (blinded, 4 lenses): unanimous RATIFY.**
- **kubernetes-specialist** (assigned dissenter): RATIFY — repo-wide hunt found zero live consumers of the removed `SEI_LOCAL`/`local` surface; sole SDK consumer (integration harness) is k8s-only; `imports_test` seam guard discriminates correctly; `SEI_LOCAL` not on the frozen one-way-door list; `provider.Register` + explicit `Open` reintroduction path open.
- **systems-engineer**: RATIFY — import-guard fails-closed and can fire (verified via `go list -deps`); `resolveMode` correct across all cases; registry concurrency untouched; drift exact-equals sound.
- **idiomatic-reviewer**: RATIFY — reads native. Advisory: fold the duplicated seam comment into the `Provider` doc.
- **prose-steward**: RATIFY (w/ reservations) — docs unambiguous. Advisory: standardize on the `provider.Register` façade (vs `RegisterProvider`) in consumer-facing instructions.

**Bugbot:** clean. **CI:** lint / test / test-integration / verify-generated all green.

The two advisory doc nits are non-blocking and will be folded into a later SDK-docs touch.

PR #496, opened 2026-07-22
feat(genesis): add Vesting field to SeiNetwork's GenesisAccount

bdchatham:
## Summary
- Adds an optional `Vesting` field (nested `GenesisAccountVesting`: `amount`, `endTime`, `delayed`) to `SeiNetwork.spec.genesis.accounts[]`. `nil`/absent is unchanged behavior — every existing caller (e.g. admin/alice/bob funding via the qa-testing nightly harness) is unaffected.
- Threads the field through every mirror point: the CRD type (`api/v1alpha1/seinetwork_types.go`, regenerated deepcopy + CRD YAML via `make generate manifests`), the plain Go SDK type (`sdk/sei/spec.go`), the k8s-provider render step (`sdk/sei/provider/k8s/render.go`), and the planner's mapping into the sidecar task request (`internal/planner/group.go`).
- Restores a coverage path lost when [sei-chain#3714](https://github.com/sei-protocol/sei-chain/pull/3714) deprecated live `MsgCreateVestingAccount`: constructing the vesting account directly at genesis-assembly time sidesteps that deprecation entirely (see companion PR below).

## ⚠️ Blocked — draft until sei-protocol/seictl#238 merges and tags
This PR is **code-complete and fully tested locally** (built + tested against seictl's branch via a temporary, uncommitted `go.mod replace` pointing at a local checkout — never pushed), but `go.mod` still pins `github.com/sei-protocol/seictl v0.0.64`, which doesn't have the new `GenesisAccountVesting` wire type / `Vesting` field that seictl#238 adds. **CI on this PR is expected to fail to build until that dependency is bumped.**

Attempted to pre-pin to seictl#238's exact commit via a pseudo-version (`go get github.com/sei-protocol/seictl@<sha>`) so this PR could build standalone, but that surfaced an unrelated, pre-existing `gogo/protobuf@v1.3.3` module-resolution error (`unknown revision v1.3.3`) once MVS re-resolved transitive versions — not something introduced by this change, and not worth fixing as a side quest here.

**Next step once seictl#238 merges and cuts a tagged release:** `go get github.com/sei-protocol/seictl@v0.0.6X && go mod tidy`, confirm clean, then this PR is ready for real review/merge.

## Test plan
- [x] `go build ./...` clean (with local seictl replace)
- [x] `go vet ./...`, `gofmt -l .` clean
- [x] `go test ./...` (excluding e2e) passes, including two new tests: `TestBuildPlan_PropagatesVesting` (planner → sidecar task JSON round-trip) and `TestRenderNetwork_PropagatesVesting` (SDK spec → CRD render), both also asserting a `nil`-Vesting sibling account stays nil, not zero-valued
- [x] `make test-integration` (full envtest suite against a live API server, validating the actual generated CRD schema) passes
- [ ] Full build against the real (non-local) seictl dependency — blocked, see above


bdchatham:
## xreview findings (5 independent reviewers: kubernetes-specialist [dissenter], solidity-developer, systems-engineer, security-specialist, idiomatic-reviewer)

Ledger verdict: **RESOLVED** — every correctness-grade finding fixed or explicitly accepted. Fixes pushed to both PRs. Summary of what changed and what's deliberately deferred:

### Fixed in code (seictl#238 @ 51396b5, this PR @ 42bf5d2)
- **Degenerate-schedule silent-unlock (3 reviewers converged — highest-signal finding).** `EndTime > 0` alone didn't guarantee a *locked* fixture: a continuous account with `EndTime <= genesisTime`, or a delayed account with a past `EndTime`, is fully unlocked from block 1 with no error (auth.InitGenesis never calls `acc.Validate()`). Now rejected at assembly time: non-zero amount, `EndTime > genesisTime`, and a belt-and-suspenders `acc.Validate()`. Tests added for all three paths.
- **Wire-mapping test gap (dissenter + idiom).** `genesisAccountsToWire`'s hand-built map had no vesting coverage; added a JSON round-trip test proving the CLI map and the server struct agree on every key.
- **CRD doc leaked internal type names (idiom S1).** `kubectl explain`-facing text now describes behavior, not `ContinuousVestingAccount`/`MsgCreateVestingAccount`.

### Deliberately accepted / deferred (with rationale)
- **⚠️ Sidecar-image skew silent-drop (dissenter, strongest finding — ACCEPTED via the lockstep constraint, not a code guard).** A controller running this PR's code against a **pre-#238 sidecar image** would have its `vesting` field silently dropped by `encoding/json` (unknown fields ignored) → a plain unlocked account, no error. This is inherent to the additive-field-over-HTTP model and is why this PR is **blocked on seictl#238 merging + tagging + a lockstep `SEITASK_IMAGE`/sidecar-image bump** (already the stated gating condition). Adding `DisallowUnknownFields` server-side would convert the silent drop into a loud error but is a **separate hardening decision** with its own blast radius (it would reject *any* forward-compatible field on the whole task API, not just this one) — flagged for a maintainer call, not bundled here.
- **Non-atomic genesis.json write (systems-engineer).** `SaveAs` truncates-then-writes with no temp+rename; pre-existing, shared by every genesis writer, not a regression from this change. Deferred.
- **Vesting-amount coin-validation is server-side, not planner-time (dissenter MISMATCH).** Rejected as a finding: the client package deliberately avoids importing the full sei-cosmos SDK (documented at the `validateSeiAccountAddress` comment), and `Balance` itself is already validated server-side only — so this follows established precedent, not a regression. The client *does* now validate the vesting sub-fields it can cheaply check (non-empty amount, positive EndTime).

Full ledger committed at `bdchatham-designs/designs/qa-testing-stack/xreview/`.

